PR #180, opened 2023-02-17
Load waveform nb

md-arif-shaikh:
example notebook for loading waveforms of different origins


vijayvarma392:
- [ ] Can we change it so that the defaults like q, chi1, chi2, ecc, meanano, omega0 and approximant are not hard coded? Just specify in these in the example you give, and mention that: "As of Feb 2023, the only eccentric model available through LAL is EccentricTD, which does not include the merger, and is restricted to nonspinning BHs. We expect more eccentric models will be added in the future." image
- [ ]  I added this 
image, which I think is really helpful; we need to make things easy for the user. But, see next point.

- [ ]  I like the idea that `get_available_waveform_origins` tells you the functions, but can we make it even easier? Like: `get_available_waveform_origins` only tells you the list of origins. And a new function, like `get_load_waveform_docs(orgin)` prints out the help text based on origin? Then, replace my above line with `#get_load_waveform_docs('LAL')` and nobody needs to know which secret function gets used for which origin.
- [ ] Add this commented-out thing for the examples of all origins to make it easy for the user.
- [ ] Always have some comments like this (please add for other origins)
image
- [ ] Having just `LAL` as the label is confusing, include the approximant name.
- [ ] Change the location to `data/SXS_BBH_XXXX/LevX/rh..` to be more helpful 
image
- [ ] Same for the lvcnr one, use the path format in lvcnr repo to be more helpful. Also, is this a Toni waveform? We should try to use a waveform that is already in lvcnr so the user can just do the same without thinking.
image

test/test_example_notebooks.py
```diff
@@ -23,8 +23,11 @@ def test_example_notebooks():
     if len(notebooks_list) == 0:
         raise Exception("No notebooks found!")
 
+    no_test_list = [f"{git_home}/examples/load_waveform_demo.ipynb"]
```
vijayvarma392, line 26: For the sake of future Arif, always explain why you are doing non-obvious things like this.

md-arif-shaikh: Done

vijayvarma392:
Remove the highlighted part, now?
image

Please do another quick scan of everything in the notebook.

gw_eccentricity/load_data.py
```diff
@@ -174,22 +194,16 @@ def load_LAL_waveform(**kwargs):
         default is "EccentricTD".
```
vijayvarma392, line 194: What about this? Please do another read of the docs in this script to make sure everything is good.

vijayvarma392: A quick scan of the code is probably good. For example, we should not be saying "Stupid EccentricTD only works for finite ecc" in public code.

examples/load_waveform_demo.ipynb
```diff
@@ -0,0 +1,507 @@
...
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "# As of Feb 2023, the only time-domain eccentric model available through LAL is EccentricTD,\n",
+    "# which does not include the merger, and is restricted to nonspinning BHs.\n",
+    "# We expect more eccentric models will be added in the future.\n",
+    "kwargs = {\n",
```
vijayvarma392, line 203: Here, also say:
Note that EccentricTD only allows ecc > 1e-5, and ignores the mean_ano parameter.

gw_eccentricity/load_data.py
```diff
@@ -524,30 +534,33 @@ def load_lvcnr_waveform(**kwargs):
         returned.
 
     include_zero_ecc: bool
-        If True returns PhenomT waveform mode for same set of parameters
+        If True, returns PhenomT waveform mode for the same set of parameters
```
vijayvarma392, line 537: PhenomT should not be hard coded like this, right?
Please make sure there are no other inconsistencies of any kind.

gw_eccentricity/load_data.py
```diff
@@ -518,26 +518,38 @@ def get_defaults_for_nr():
 def load_lvcnr_waveform(**kwargs):
     """Load modes from lvcnr files.
 
+    Loading waveform modes from files in lvcnr format require the file
+    `SEOBNRv4ROM_v2.0.hdf5` in `LAL_DATA_PATH`.  NR simulation data is
+    compressed using ROM and saved in LVCNR format to reduce file size and the
+    above file is required to correctly interpolate the NR data while loading
+    the waveform modes.
```
vijayvarma392, line 525: Not quite right, I think it's also used to get the peak time for some reason.
Just simplify and say the same thing you say for zeroecc: Needs so and so file, get it from here, set LAL_DATA_PATH, etc.

gw_eccentricity/load_data.py
```diff
@@ -518,26 +518,38 @@ def get_defaults_for_nr():
...
-        If Momega0 = 0, uses the entire NR data. The actual Momega0 will be
-        returned.
+        If Momega0 = 0, uses the entire NR data.
 
     include_zero_ecc: bool
-        If True, returns PhenomT waveform mode for the same set of parameters
-        except eccentricity set to zero.
+        If True, returns zero eccentricity waveform mode (only (2, 2) mode) for
+        the same set of parameters except eccentricity set to zero.
+        The zero eccentricity waveform is generated using a waveform model
```
vijayvarma392, line 551: a waveform model --> the waveform model

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
md-arif-shaikh: Done

gw_eccentricity/load_data.py
md-arif-shaikh: Done

gw_eccentricity/load_data.py

vijayvarma392 (approved):

PR #186, opened 2023-04-29
Sphinx doc

md-arif-shaikh:
Create docs automatically and host on readthedocs


docs/source/gw_eccentricity.rst
```diff
@@ -0,0 +1,117 @@
+gw\_eccentricity package
+========================
```
vijayvarma392, line 2: Questions: Is this file autogenerated? Because otherwise, it would be hell to maintain this whenever there are new functions, etc. And if it is autogenerated:
1. Does it need to be in the repo, or will readdthedocs generate it itself.
2. Whenever there is a change in gw_ecc, what command needs to be run to generate these?

md-arif-shaikh: remove that file. It can be autogenerated in the pre-build step. Added that step in `.readthedocs.yaml`

vijayvarma392: Ok, please go through the [Files changed](https://github.com/vijayvarma392/gw_eccentricity/pull/186/files) and confirm that there are no other files that should not go in the repo. 

This can be accessed here: 
image

It's important to look at this for all PRs, so you catch issues that may have slipped you attention when writing the code. Also, save me review effort :)

md-arif-shaikh: checked. No other unnecessary files.

vijayvarma392 (approved):

PR #190, opened 2023-06-16
fix colormaps import for older mpl

md-arif-shaikh:


gw_eccentricity/plot_settings.py
```diff
@@ -1,9 +1,15 @@
 """Fancy settings for plots."""
 from matplotlib import rc
 from cycler import cycler
-from matplotlib import colormaps
 
-dark2 = colormaps["Dark2"].colors
+# Getting colormaps in matplotlib has changed from v3.5.0
+import matplotlib
+if matplotlib.__version__ < "3.5.0":
```
vijayvarma392, line 7: Does this always work?
Is something in this discussion relevant for this: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python?

md-arif-shaikh: Seems like using `packaging` would be better. We need this dependency in our package in that case. Is that okay?

vijayvarma392: Yeah, unclear if there is clear winning strategy :).
I suggest maybe some googling to see what people think, but whichever way you think is best is fine with me.

md-arif-shaikh: I have used `packaging` to compare the versions

vijayvarma392 (approved):

vijayvarma392:
Please make sure to close the issue with a comment there.

Issue #194, opened 2023-06-19
add readmedocs badge in the readme


md-arif-shaikh:

md-arif-shaikh:
Addressed in #196

PR #196, opened 2023-06-22
Add docs badge link, addresses #194

duetosymmetry:


md-arif-shaikh:
The patch itself looks good, however what about the build not running when PR is from a fork instead of a branch? @vijayvarma392

vijayvarma392:
> The patch itself looks good, however what about the build not running when PR is from a fork instead of a branch? @vijayvarma392

Ok, feel free to merge, but:
Please leave a comment explaining the reason, for Leo's sake.
Update the instructions here, so we say we prefer the branch from main repo + PR approach: https://github.com/vijayvarma392/gw_eccentricity/blob/main/README_developers.md#contributing-to-gw_eccentricity.

duetosymmetry:
Wait, are you saying that if somebody wants to contribute, you have to first give them write access to your repo so they can push a branch?

md-arif-shaikh:
Yeah, I don't know if it's the correct approach, can't we figure out what workflow would work for fork + PR?

vijayvarma392:
> Yeah, I don't know if it's the correct approach, can't we figure out what workflow would work for fork + PR?

Sure, would be nice to fix it.
It can be turned on for PRs like this: https://github.com/vijayvarma392/gw_eccentricity/commit/57b184c50379f7a383277131f663038cd0b1c944, but this was running tests twice in some scenarios that I don't remember anymore.

md-arif-shaikh:
> > Yeah, I don't know if it's the correct approach, can't we figure out what workflow would work for fork + PR?
> 
> Sure, would be nice to fix it. It can be turned on for PRs like this: [57b184c](https://github.com/vijayvarma392/gw_eccentricity/commit/57b184c50379f7a383277131f663038cd0b1c944), but this was running tests twice in some scenarios that I don't remember anymore.

Maybe we should turn that on and see if there are issues. I looked at a few public GitHub projects and the workflow is on both push and pull_request. So maybe we will not be doing anything totally wrong.

md-arif-shaikh:
Hi @duetosymmetry, the workflow is modified so that tests run for PR from forks. Could you pull the changes from main so that it passes the tests before we merge this?

md-arif-shaikh (approved):

Issue #60, opened 2022-05-28
dataDict should know how to handle cases where amp/phase are provided directly


vijayvarma392:
There are situations where models already have access to Amp22 and phi22. We should not force them to first compute h22. Instead, if Amp22/phi22 are in dataDict, it should just use them. Similarly, if omega22 is also provided, it should not recompute it.

md-arif-shaikh:
implemented in #197

PR #197, opened 2023-06-27
allow Amp, omega, phase in dataDict

md-arif-shaikh:
Addresses #60. Allows providing `amplm`, `omegalm`, `phaselm` in `dataDict`. If all of these or a subset of these are provided, new functions are added to prioritize which data to use.

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -443,7 +443,7 @@ class eccDefinition:
                                        dataDict,
                                        num_orbits_to_exclude_before_merger,
                                        extra_kwargs):
-        """Truncate dataDict if "num_orbits_to_exclude_before_merger" is not None.
+        """Truncate dataDict if `num_orbits_to_exclude_before_merger` is not None.
 
         parameters
         ----------
@@ -456,9 +456,14 @@ class eccDefinition:
 
         Returns
         -------
-        dataDict:
-            Truncated if num_orbits_to_exclude_before_merger is not None
-            else the unchanged dataDict.
+        newDataDict:
+            Dictionary containing the amplitude, phase and omega of the
+            eccentric waveform modes. Includes the same of the zeroecc waveform
+            modes when present in the input `dataDict`.
+
+            If `num_orbits_to_exclude_before_merger` is not None, then the
+            eccentric data, i.e., amplitude, phase, omega of the available
+            modes, are truncated before newDataDict is returned.
         t_merger:
             Merger time evaluated as the time of the global maximum of
             amplitude_using_all_modes. This is computed before the truncation.
@@ -522,14 +527,11 @@ class eccDefinition:
                     num_orbits_to_exclude_before_merger)
             # Truncate amp, phase, omega in eccentric waveform data.
             for k in ["amplm", "phaselm", "omegalm"]:
-                if k in newDataDict:
-                    for mode in newDataDict[k]:
-                        newDataDict[k][mode] \
-                            = newDataDict[k][mode][
-                                :index_num_orbits_earlier_than_merger]
-                        newDataDict["t"] \
-                            = newDataDict["t"][
-                                :index_num_orbits_earlier_than_merger]
+                for mode in newDataDict[k]:
+                    newDataDict[k][mode] = newDataDict[k][mode][
+                        :index_num_orbits_earlier_than_merger]
+                    newDataDict["t"] = newDataDict["t"][
+                        :index_num_orbits_earlier_than_merger]
         return newDataDict, t_merger, amp22_merger, min_width_for_extrema
 
     def get_width_for_peak_finder_from_phase22(self,
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -44,11 +44,13 @@ class eccDefinition:
                         "omegalm_zeroecc": omegaDict,
                        }
 
-            "t" and at least one of the followings are mandatory:
+            "t" and one of the followings are mandatory:
 
             - "hlm"
             - "amplm" and "phaselm"
 
+            Note that providing "hlm", "amplm" and "phaselm" all at the same
+            time will raise error.
             Apart from specifying "hlm" or "amplm" and "phaselm", the user can
             also provide "omegalm". If the "omegalm" key is not explicitly
             provided, it is computed from the given "hlm" or "phaselm" using
@@ -56,11 +58,13 @@ class eccDefinition:
 
             The keys with suffix "zeroecc" are only required for
             `ResidualAmplitude` and `ResidualFrequency` methods, where
-            "t_zeroecc" and at least one of the followings are to be provided:
+            "t_zeroecc" and one of the followings are to be provided:
 
             - "hlm_zeroecc"
             - "amplm_zeroecc" and "phaselm_zeroecc"
 
+            Note that providing "hlm_zeroecc", "amplm_zeroecc" and
+            "phaselm_zeroecc" all at the same time will raise error.
             Similar to "omegalm", the user can also provide "omegalm_zeroecc".
             If it is not provided in `dataDict`, it is computed from the given
             "hlm_zeroecc" or "phaselm_zeroecc" using finite difference method.
@@ -506,6 +510,15 @@ class eccDefinition:
             Minimum width for the `find_peaks` function. This is computed
             before the truncation.
         """
+        # Check that the dataDict contains either the hlms or the amplm/phaselm
+        # and not both of these.
+        for suffix in ["", "_zeroecc"]:
+            if all(key+suffix in dataDict for key in
+                   ["hlm", "amplm", "phaselm"]):
+                raise Exception(
+                    f"Provide either `hlm{suffix}` or both of `amplm{suffix}` "
+                    f"and `phaselm{suffix}`. `dataDict` should not contain "
+                    "all of the three at the same time.")
         # Create a new dictionary that will contain the data necessary for
         # eccentricity measurement.
         newDataDict = {}
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -49,7 +49,6 @@ class eccDefinition:
 
             - "hlm"
             - "amplm" and "phaselm"
-            - "amplm" and "omegalm".
 
             The keys with suffix "zeroecc"
             are only required for `ResidualAmplitude` and
@@ -58,7 +57,6 @@ class eccDefinition:
 
             - "hlm_zeroecc"
             - "amplm_zeroecc" and "phaselm_zeroecc"
-            - "amplm_zeroecc" and "omegalm_zeroecc".
 
             If provided for other methods, they are
             used for additional diagnostic plots, which can be helpful
@@ -69,14 +67,14 @@ class eccDefinition:
 
             - "t": 1d array of times.
 
-                - Should be uniformly sampled, with a small enough time step
-                  so that omega22(t) can be accurately computed. We use a
-                  4th-order finite difference scheme. In dimensionless units,
-                  we recommend a time step of dtM = 0.1M to be conservative,
-                  but one may be able to get away with larger time steps like
-                  dtM = 1M. The corresponding time step in seconds would be dtM
-                  * M * lal.MTSUN_SI, where M is the total mass in Solar
-                  masses.
+                - Should be uniformly sampled, with a small enough time step so
+                  that omega22(t) can be accurately computed, if necessary. We
+                  use a 4th-order finite difference scheme. In dimensionless
+                  units, we recommend a time step of dtM = 0.1M to be
+                  conservative, but one may be able to get away with larger
+                  time steps like dtM = 1M. The corresponding time step in
+                  seconds would be dtM * M * lal.MTSUN_SI, where M is the total
+                  mass in Solar masses.
                 - We do not require the waveform peak amplitude to occur at any
                   specific time, but tref_in should follow the same convention
                   for peak time as "t".
@@ -101,13 +99,13 @@ class eccDefinition:
               with "t". Should have the same format as "hlm", except that the
               phase is real. The phaselm is related to hlm as hlm = amplm *
               exp(- i phaselm) ensuring that the phaselm is monotonically
-              increasing.
+              increasing for m > 0 modes.
 
             - "omegalm": Dictionary of the frequencies of the waveform modes
               associated with "t". Should have the same format as "hlm", except
               that the omegalm is real. omegalm is obtained from the phaselm
               (see above) as omegalm = d/dt phaselm, which means that the
-              omegalm is positive.
+              omegalm is positive for m > 0 modes.
 
             - "t_zeroecc" and "hlm_zeroecc":
 
@@ -129,7 +127,6 @@ class eccDefinition:
             - "amplm_zeroecc", "phaselm_zeroecc" and "omegalm_zeroecc":
                 Same as "amplm", "phaselm" and "omegalm", respectively, but
                 for the quasicircular counterpart to the eccentric waveform.
-                Should have same format as "hlm_zeroecc".
 
         num_orbits_to_exclude_before_merger:
                 Can be None or a non negative number.  If None, the full
@@ -231,7 +228,7 @@ class eccDefinition:
         """
         # Get data necessary for eccentricity measurement
         self.dataDict, self.t_merger, self.amp22_merger, \
-            min_width_for_extrema = self.get_necessary_data(
+            min_width_for_extrema = self.process_data_dict(
                 dataDict, num_orbits_to_exclude_before_merger, extra_kwargs)
         self.t = self.dataDict["t"]
         # check if the time steps are equal, the derivative function
@@ -330,7 +327,7 @@ class eccDefinition:
         For example, if `dataDict` has only the complex hlm modes, this
         functions uses the hlm modes to create a dictionary containing the
         amplitude, phase and omega by decomposing the hlm modes. Afterwards,
-        only these waveform data in the new dataDict is used for all the
+        only these waveform data in the new dataDict is used for all
         further computations.
         """
         amp_phase_omega_dict = {}
@@ -360,8 +357,6 @@ class eccDefinition:
                 {"omegalm_zeroecc": self.get_omegalm_from_phaselm(
                     dataDict["t_zeroecc"],
                     amp_phase_omega_dict["phaselm_zeroecc"])})
-        else:
-            pass
         return amp_phase_omega_dict
 
     def get_amplm_from_dataDict(self, dataDict):
@@ -379,13 +374,10 @@ class eccDefinition:
                 # compute amplitude of each hlm mode and add to ampDict
                 for k in dataDict["hlm" + suffix]:
                     ampDict.update({k: np.abs(dataDict["hlm" + suffix][k])})
-            else:
-                if suffix == "_zeroecc":
-                    # zeroecc amplitude is required only for Residual methods.
-                    pass
-                else:
-                    raise Exception("dataDict should contain at least one of"
-                                    " `'amplm'` or `hlm` dict.")
+            elif suffix == "":
+                raise Exception("dataDict should contain at least one of "
+                                "['amplm' 'hlm'] for computing amplitude of "
+                                "the waveform modes.")
         amplmDict = {"amplm": amplm}
         if amplm_zeroecc:
             amplmDict.update({"amplm_zeroecc": amplm_zeroecc})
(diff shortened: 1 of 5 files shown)


gw_eccentricity/eccDefinition.py
```diff
@@ -192,12 +234,15 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
             raise Exception("Input time array must have uniform time steps.\n"
                             f"Time steps are {self.t_diff}")
-        self.hlm = self.dataDict["hlm"]
-        self.h22 = self.hlm[(2, 2)]
-        self.amp22 = np.abs(self.h22)
-        self.phase22 = - np.unwrap(np.angle(self.h22))
-        self.omega22 = time_deriv_4thOrder(self.phase22,
-                                           self.t[1] - self.t[0])
+        # get phase22 from dataDict
+        self.phase22, self.omega22, self.h22 = self.compute_phase22(self.dataDict)
```
vijayvarma392, line 238: Well, it does more than compute phase...
Maybe a single function called decompose_waveform, which returns phase22, omega22, amp22 and h22?

gw_eccentricity/eccDefinition.py
```diff
@@ -192,12 +234,15 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+        # get amp22 from dataDict
+        self.amp22, _ = self.compute_amp22(self.dataDict)
+        # get omega22 from dataDict
+        # check if omega22 has been already computed when getting
+        # phase22.
+        if self.omega22 is None:
+            self.omega22 = self.compute_omega22(self.dataDict)
```
vijayvarma392, line 245: Just do this in that one function?

gw_eccentricity/eccDefinition.py
```diff
@@ -251,16 +296,25 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         self.t_for_omega22_average = None
         self.omega22_average = None
 
-        if "hlm_zeroecc" in self.dataDict:
-            self.compute_res_amp_and_omega22()
+        # compute residual data
+        if any(["hlm_zeroecc" in self.dataDict,
+                "amplm_zeroecc" in self.dataDict and "omegalm_zeroecc" in self.dataDict,
+                "amplm_zeroecc" in self.dataDict and "phaselm_zeroecc" in self.dataDict]):
+            self.compute_res_amp22_and_res_omega22()
```
vijayvarma392, line 303: Why does this need a separate function?

gw_eccentricity/eccDefinition.py
```diff
@@ -251,16 +296,25 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         self.t_for_omega22_average = None
         self.omega22_average = None
 
-        if "hlm_zeroecc" in self.dataDict:
-            self.compute_res_amp_and_omega22()
+        # compute residual data
+        if any(["hlm_zeroecc" in self.dataDict,
+                "amplm_zeroecc" in self.dataDict and "omegalm_zeroecc" in self.dataDict,
+                "amplm_zeroecc" in self.dataDict and "phaselm_zeroecc" in self.dataDict]):
```
vijayvarma392, line 302: Just check for `hlm_zeroecc` and `amplm_zeroecc`? Why do you care here whether phase or omega is given?

gw_eccentricity/eccDefinition.py
```diff
@@ -295,19 +349,20 @@ def truncate_dataDict_if_necessary(self,
             truncation.
         """
         t = dataDict["t"]
-        phase22 = - np.unwrap(np.angle(dataDict["hlm"][(2, 2)]))
+        # Get phase22 depending on what is provided in dataDict
+        phase22, omega22, h22 = self.compute_phase22(dataDict)
```
vijayvarma392, line 353: It's not nice that you have to do this here, and again after truncation.
Is there a better way? This is also a bit confusing.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: I agree with using a better name. But putting everything in the same function makes it too long. Also, we do not need to compute each of these data at the same time. For example, in the truncate_dataDict.., we just need the phase and amp, we do not need h22 or omega22 there.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Do you mean putting these lines for getting amp22, phase22, and omega22 in a single wrapper function and calling it? As mentioned above, I don't think writing a single function that gets all these is better than keeping separate functions for each of these.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: What do you mean? computing the residual amp22 and omega22 is lengthy, we did not want to do all those staff in init itself and decided to do it in a separate function and call it in init.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Because we can compute residual omega22 if omega22_zeroecc is provided? Note: We are computing both omega22 residual and amp22 residual in the following function.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: We discussed this earlier. The thing is that we need information about the full amp22 and phase22 to set the minimum width for peak_finding before truncating the dataDict. Once it passes the truncated dataDict, we use that to get waveform quantities to be used for extrema finding.

gw_eccentricity/eccDefinition.py
```diff
@@ -67,7 +93,18 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                     m) waveform mode. Should contain at least the (2, 2) mode,
                     but more modes can be included, as indicated by the
                     ellipsis '...'  above.
+
+            - "amplm": Dictionary of amplitudes of waveform modes associated
+              with "t". Should have the same format as "hlm".
```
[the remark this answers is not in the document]
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
```diff
@@ -31,21 +32,44 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                        "amplm_zeroecc": ampDict,
+                        "phaselm_zeroecc": phaseDict,
+                        "omegalm_zeroecc": omegaDict,
+                       }
+
+            "t" and at least one of the followings are mandatory:
+
+            - "hlm"
+            - "amplm" and "phaselm"
+            - "amplm" and "omegalm".
```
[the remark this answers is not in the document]
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
```diff
@@ -251,24 +297,178 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+        eccentricity and mean anomaly, we need amplitude, phase and omega. In
+        this function, we get these data from `dataDict` and create a new
+        dictionary that contains only the amplitude, phase and omega of the
+        available modes.
+
+        For example, if `dataDict` has only the complex hlm modes, this
+        functions uses the hlm modes to create a dictionary containing the
+        amplitude, phase and omega by decomposing the hlm modes. Afterwards,
+        only these waveform data in the new dataDict is used for all the
+        further computations.
```
vijayvarma392, line 334: "all the further computations" --> "all further computations"
Good to run a grammar check.

gw_eccentricity/eccDefinition.py
```diff
@@ -251,24 +297,178 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+            elif "hlm" + suffix in dataDict:
+                # Compute phase of each mode in hlm and add to phaselm
+                for k in dataDict["hlm" + suffix]:
+                    phaseDict.update(
+                        {k: - np.unwrap(
+                            np.angle(dataDict["hlm" + suffix][k]))})
+            elif "omegalm" + suffix in dataDict:
+                # Compute phase of each mode from omega and add to phaseDict
+                for k in dataDict["omegalm" + suffix]:
+                    phaseDict.update(
```
vijayvarma392, line 416: This I'm not sure about.
It might be better to require:
Provide either:
1. hlm
2. Alm, philm
3. Alm, philm, omegalm
Instead of doing the integration under the hood, it's probably best to leave it to the user. Most waveforms have easy access to the phase, so having omega and not phase is unusual. My worry is that users that might have the phase might forget about it and just pass omega. And cumtrapz is only second-order accurate, so could eventually become a problem. The reason for allowing omega is the user may have a better omega than we get using a time derivative, in which case, we might as well use that.

gw_eccentricity/eccDefinition.py
```diff
@@ -251,24 +297,178 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+        # Look for zeroecc phaselm is amp_phase_omega_dict and compute
+        # omegalm_zeroecc from it if phaselm_zeroecc is present
+        elif "phaselm_zeroecc" in amp_phase_omega_dict \
+             and "t_zeroecc" in dataDict:
+            amp_phase_omega_dict.update(
+                {"omegalm_zeroecc": self.get_omegalm_from_phaselm(
+                    dataDict["t_zeroecc"],
+                    amp_phase_omega_dict["phaselm_zeroecc"])})
+        else:
+            pass
```
vijayvarma392, line 364: Why do you need this else condition if you're just going to do pass?

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
```diff
@@ -31,21 +31,44 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
             used for additional diagnostic plots, which can be helpful
             for all methods. Any other keys in dataDict will be
             ignored, with a warning.
 
             The recognized keys are:
+
             - "t": 1d array of times.
+
                 - Should be uniformly sampled, with a small enough time step
                   so that omega22(t) can be accurately computed. We use a
```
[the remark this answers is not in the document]

gw_eccentricity/eccDefinition.py
```diff
@@ -67,7 +92,25 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+
+            - "amplm": Dictionary of amplitudes of waveform modes associated
+              with "t". Should have the same format as "hlm", except that the
+              amplitude is real.
+
+            - "phaselm": Dictionary of phases of waveform modes associated
+              with "t". Should have the same format as "hlm", except that the
+              phase is real. The phaselm is related to hlm as hlm = amplm *
+              exp(- i phaselm) ensuring that the phaselm is monotonically
+              increasing.
```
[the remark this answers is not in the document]
md-arif-shaikh: Corrected

gw_eccentricity/eccDefinition.py
md-arif-shaikh: done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Changed the code. It now requires the user to provide either hlm or amp and phase. If for example, amp and only omega are provided it will raise exception.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: my bad. I assumed that I need to end the if-else always with an else, but apparently, it can just end with an elif. Removed it now.

vijayvarma392 (changes requested):

vijayvarma392:
I think this is almost there. You mentioned a test for the different dataDict options. I think that's a good idea, and recommend doing it in this PR itself.

gw_eccentricity/eccDefinition.py
```diff
@@ -31,34 +31,57 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
-            "t" and "hlm" are mandatory. "t_zeroecc" and "hlm_zeroecc"
+                        "amplm_zeroecc": ampDict,
+                        "phaselm_zeroecc": phaseDict,
+                        "omegalm_zeroecc": omegaDict,
+                       }
+
+            "t" and at least one of the followings are mandatory:
+
+            - "hlm"
+            - "amplm" and "phaselm"
```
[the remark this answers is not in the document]
md-arif-shaikh: I can be provided with both `hlm` and `amplm/phaselm`. Added to doc.

gw_eccentricity/eccDefinition.py
```diff
@@ -67,7 +90,25 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+
+            - "phaselm": Dictionary of phases of waveform modes associated
+              with "t". Should have the same format as "hlm", except that the
+              phase is real. The phaselm is related to hlm as hlm = amplm *
+              exp(- i phaselm) ensuring that the phaselm is monotonically
+              increasing for m > 0 modes.
+
+            - "omegalm": Dictionary of the frequencies of the waveform modes
+              associated with "t". Should have the same format as "hlm", except
+              that the omegalm is real. omegalm is obtained from the phaselm
```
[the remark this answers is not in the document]

test/test_dataDict.py
```diff
@@ -0,0 +1,127 @@
...
+                  "ecc": 0.1,
+                  "mean_ano": 0,
+                  "include_zero_ecc": True}
+    dataDict = load_data.load_waveform(**lal_kwargs)
+
+    # get amplm, phaselm, omegalm
+    phaselm = {}
+    amplm = {}
+    omegalm = {}
+    dt = dataDict["t"][1] - dataDict["t"][0]
```
vijayvarma392, line 28: Add a sanity check that the time array is uniform. Good to add checks like this in multiple places, so that when we eventually switch to nonuniform time arrays, we know which places changes need to be made. 

In general, important to think about what your code is assuming, and add sanity checks -- I always try to do the exercise of "How can I break my code?".

test/test_dataDict.py
```diff
@@ -0,0 +1,127 @@
...
+        "t": dataDict["t"],
+        "amplm": amplm,
+        "phaselm": phaselm}
+
+    for method in available_methods:
+        # Using hlm
+        if "Residual" in method:
+            # add zeroecc hlm
+            data_hlm.update({"t_zeroecc": dataDict["t_zeroecc"],
+                             "hlm_zeroecc": dataDict["hlm_zeroecc"]})
```
vijayvarma392, line 62: Why not just do this in line 49 itself?

test/test_dataDict.py
```diff
@@ -0,0 +1,127 @@
...
+        return_dict_hlm = measure_eccentricity(
```
vijayvarma392, line 63: Confusing name...the retuned quantity is not a waveform dict. How about result_with_hlm?

test/test_dataDict.py
```diff
@@ -0,0 +1,127 @@
...
+            tref_in=dataDict["t"],
+            method=method,
+            dataDict=data_hlm)
+        gwecc_object_hlm = return_dict_hlm["gwecc_object"]
+        eccentricity_w_hlm = gwecc_object_hlm.eccentricity
+        mean_anomaly_w_hlm = gwecc_object_hlm.mean_anomaly
```
vijayvarma392, line 69: Be consistent. Sometimes you use `_hlm` sometimes `_w_hlm`. I like `_w_hlm` instead of my suggestion `with_hlm` above.

test/test_dataDict.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Using hlm with omegalm also provided
+        data_hlm.update({"omegalm": omegalm})
+        if "Residual" in method:
+            data_hlm.update(
+                {"omegalm_zeroecc": omegalm_zeroecc})
```
vijayvarma392, line 75: Same, why not just do this in line 72?

vijayvarma392 (changes requested):

test/test_dataDict.py
md-arif-shaikh: Done

test/test_dataDict.py
md-arif-shaikh: The residual data are not required for methods other than Residual ones. In most cases, users may try the non-residual methods first. Thus, I don't include the zeroecc data when the method is not Residual to make sure that there is no hidden dependence on zeroecc data in non-residual methods.

test/test_dataDict.py
md-arif-shaikh: Done

test/test_dataDict.py
md-arif-shaikh: Done

test/test_dataDict.py
md-arif-shaikh: as explained above

test/test_dataDict.py
```diff
@@ -26,6 +26,14 @@ def test_dataDict():
...
     omegalm = {}
     dt = dataDict["t"][1] - dataDict["t"][0]
+    # check that the time array is uniform. This is required to obtain omega by
+    # taking derivative of phase applying time_deriv_4thorder.
+    ids = np.where(np.abs(np.diff(dataDict["t"]) - dt) > 1e-7)[0]
+    if len(ids) > 0:
+        dt_ = np.diff(dataDict["t"])[ids[0]]
+        raise Exception("t_zeroecc is not uniform. dt at the start of the "
+                        f"time array is {dt}. dt={dt_} between {ids[0]} and "
+                        f"{ids[0]+1}. The difference is {dt - dt_}")
```
vijayvarma392: You already did a better version of this check here: https://github.com/vijayvarma392/gw_eccentricity/blob/main/gw_eccentricity/eccDefinition.py#L192, just repeat it here? Or make it a function if you like.

vijayvarma392 (changes requested):

test/test_dataDict.py
vijayvarma392: But, if I understand correctly, once you do this update, zeroecc data will be used for all methods, including nonresidual ones. Doesn't that defeat the purpose of what you are saying? Unless you get lucky and the order of the methods is so that the Residual ones happen at the end. But then, if someone creates a new method with some random name, that won't be covered in this scheme.

How about the much simpler: 
Make a function that takes two data_dicts: data_dict_hlm and data_dict_other.
This function compares the output of the two, and you just call this function multiple times with different dicts as data_dict_other.
Do the whole exercise with data_dicts with and without zero_ecc data. When the function tries to test it for a Residual method, but the data_dict doesn't contain residual data, just make it `continue` in the loop.
Something like that would make the code much easier to follow, and with far fewer long variable names. 
Added bonus, you test the nonresidual methods with and without zeroecc data, which are both realistic scenarios for a user.
What say?

test/test_dataDict.py
md-arif-shaikh: Done

test/test_dataDict.py
md-arif-shaikh: Okay, did change the code. I think it now checks more combinations of keys than I was doing earlier.

gw_eccentricity/eccDefinition.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
             - "hlm"
             - "amplm" and "phaselm"
 
-            Note that providing "hlm", "amplm" and "phaselm" all at the same
+            Note that providing "hlm" and "amplm" or "phaselm" at the same
```
vijayvarma392, line 52: Very confusing, make the whole thing simpler:
and one of the followings are mandatory:
 - "hlm" OR
 - "amplm" and "phaselm"
 but not both at the same time.

vijayvarma392: where I also did "one of the followings are" --> "one of the following is"

vijayvarma392: I edited my suggestion, please take a look now.

gw_eccentricity/eccDefinition.py
```diff
@@ -513,12 +513,13 @@ def process_data_dict(self,
...
-            if all(key+suffix in dataDict for key in
-                   ["hlm", "amplm", "phaselm"]):
+            if "hlm"+suffix in dataDict and ("amplm"+suffix in dataDict
+                                             or "phaselm"+suffix in dataDict):
                 raise Exception(
                     f"Provide either `hlm{suffix}` or both of `amplm{suffix}` "
                     f"and `phaselm{suffix}`. `dataDict` should not contain "
-                    "all of the three at the same time.")
+                    f"`hlm{suffix}` and `amplm{suffix}` or `phaselm{suffix}` "
+                    "at the same time.")
```
vijayvarma392, line 522: If I understand correctly, this won't raise an exception if I give only amplm in dataDict, is that correct?
Please always test code like this by making up examples.

vijayvarma392: You probably want do an xor, which is `^` in python.
Maybe something like
if hlm is there ^ (amplm is there and philm is there):
      raise Exception
but you should test it.

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Okay, noticed the other comment. will look at the `^` thing.

gw_eccentricity/eccDefinition.py
```diff
@@ -508,16 +508,28 @@ def process_data_dict(self,
...
-                    "at the same time.")
+                    f"`dataDict` {'should' if suffix == '' else 'may'} "
+                    "contain one of the following: \n"
+                    f"1. 'hlm{suffix}' OR \n"
+                    f"2. 'amplm{suffix}' and 'phaselm{suffix}'\n"
+                    "But not both 1. and 2. at the same time.")
+        # Raise exception if hlm is not provided but either amplm and/or
+        # phaselm is also not provided
+        if ("hlm" not in dataDict) and any(["amplm" not in dataDict,
+                                            "phaselm" not in dataDict]):
```
vijayvarma392, line 528: BTW, seeing this, it seems unnessarily convoluted.
Can't you just do:
`all(k in dataDict for k in ["hlm", "amplm", "phaselm"])`
And similar for the other any above?

vijayvarma392 (approved):

Issue #84, opened 2022-06-17
How should we handle failures?


vijayvarma392:
What should the code do when it cannot measure ecc?
For example, for the Amp method when we are at small ecc and very close to the merger.
When tref_in has a bunch of times/freqs, it just returns those times/freqs where it can measure ecc, but think of the PE scenario where the user has one time/freq and sometimes that happens to not work.

How about the following behavior: 
By default, it raises an error.
But if an option like "failures_are_set_to_zero" = True, the method then sets ecc=mean_ano=0.
There should be a big warning in the docs about using this.

But, at least for the Amp method, we expect failures only when ecc is really small, so this might be good enough at current sensitivities. I mean, who cares about the difference between ecc=1e-5 and 1e-3?

One would have to be more careful for FrequencyFits, of course, at least until we are happy with its robustness.

md-arif-shaikh:
Moving this to #206

PR #203, opened 2023-08-24
set failures to zero

md-arif-shaikh:


md-arif-shaikh:
Addresses #84

gw_eccentricity/eccDefinition.py
```diff
@@ -237,6 +238,12 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                 methods. See
                 eccDefinitionUsingFrequencyFits.get_default_kwargs_for_fits_methods
                 for allowed keys.
+
+            set_failures_to_zero : bool, default=False
+                If True and the waveform is sufficiently long then instead of
+                raising exception when number of extrema is insufficient to
+                build frequency interpolant through the extrema, eccentricity
+                and mean anomaly are set to zero.
```
vijayvarma392, line 246: Unclear, and not scary enough, how about:
The code normally raises an exception if sufficient number of extrema are not found. This can happen for various reasons including when the eccentricity is too small for some methods (like the Amplitude method) to measure. See e.g. Fig.XX of arxiv.xxxx. If set_failures_to_zero is set to True, we assume that small eccentricity is the cause, and set the returned eccentricity and mean anomaly to zero when sufficient extrema are not found. USE THIS WITH CAUTION!

gw_eccentricity/eccDefinition.py
```diff
@@ -305,6 +313,13 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         # called, these get set in that function.
         self.t_for_omega22_average = None
         self.omega22_average = None
+        # Approximate number of orbits derived using the phase of 22 mode
+        # assuming that a phase change 4pi occurs over an orbit.
+        self.approximate_num_orbits = ((self.phase22[-1] - self.phase22[0])
+                                       / (4 * np.pi))
+        # The following is updated to True when the waveform has enough number
+        # of orbits but a method can not find sufficient number of extrema.
+        self.probably_quasicircular = False
```
vijayvarma392, line 322: Does this need to be in self? Why not just in the function where this is determined?

gw_eccentricity/eccDefinition.py
```diff
@@ -1055,26 +1071,41 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+                    "zero.",
+                    important=True,
+                    debug_level=0)
             else:
-                method_message = ""
-            raise Exception(
-                f"Number of {extrema_type} found = {num_extrema}.\n"
-                f"Can not build frequency interpolant through the {extrema_type}.\n"
-                f"{method_message}")
+                recommended_methods = ["ResidualAmplitude", "AmplitudeFits"]
```
vijayvarma392, line 1098: Why not ResidualFrequency as well?

gw_eccentricity/eccDefinition.py
```diff
@@ -1055,26 +1071,41 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+                if self.method not in recommended_methods:
+                    method_message = (
+                        "It's possible that the eccentricity is too "
+                        f"low for the {self.method} method to detect"
+                        f" the {extrema_type}. Try one of "
+                        f"{recommended_methods}.")
+                else:
+                    method_message = ""
+                    raise InsufficientExtrema(extrema_type, num_extrema,
+                                              method_message)
```
vijayvarma392, line 1108: Wait, you only raise an Exception for recommended_method?

gw_eccentricity/eccDefinition.py
```diff
@@ -1372,6 +1415,39 @@ def measure_ecc(self, tref_in=None, fref_in=None):
             return_dict.update({"tref_out": self.tref_out})
         return return_dict
 
+    def set_eccentricity_and_mean_anomaly_to_zero(
+            self, tref_in, fref_in):
+        """Set eccentricity and mean_anomaly to zero."""
+        return_dict = {}
+        if tref_in is not None:
+            # check that tref_in is in the allowed range
+            ndim = np.ndim(tref_in)
```
vijayvarma392, line 1424: Is this code copied from somewhere else? If so, don't copy, make a function.

gw_eccentricity/load_data.py
```diff
@@ -261,6 +261,9 @@ def load_LAL_waveform(**kwargs):
             zero_ecc_kwargs['ecc'] = 0
         zero_ecc_kwargs['include_zero_ecc'] = False   # to avoid infinite loops
         dataDict_zero_ecc = load_waveform(**zero_ecc_kwargs)
+        while dataDict_zero_ecc['t'][0] > dataDict["t"][0]:
```
vijayvarma392, line 264: Add a comment about what you are doing.

gw_eccentricity/eccDefinition.py
```diff
@@ -1237,6 +1279,17 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
         original_apocenters = apocenters.copy()
         self.check_num_extrema(apocenters, "apocenters")
+
+        # If the eccentricity is too small for a method to find the extrema and
+        # set_failures_to_zero is set to true, then we set the eccentricity and
+        # mean anomaly to zero and return it.
+        # In this case, the rest of the code in this function is not executed and
+        # therefore, many variables which are used in diagnostic tests are never
+        # computed thus making diagnostics irrelevant.
+        if self.probably_quasicircular and self.set_failures_to_zero:
```
vijayvarma392, line 1289: If this is the reason you have probably_quasicircular in self, it is much clearer to have probably_quasicircular returned by self.check_num_extrema. Otherwise, it's very unclear where what gets set.

gw_eccentricity/exceptions.py
```diff
@@ -0,0 +1,58 @@
+"""Custom exception classes for gw_eccentricity."""
+
+
+class InsufficientExtrema(Exception):
```
vijayvarma392, line 4: I don't get why you need this so far. Please explain in the comments why you need this, for the sake of future you, if nothing else.

gw_eccentricity/exceptions.py
```diff
@@ -0,0 +1,58 @@
...
+        self.additional_message = additional_message
+        self.message = (f"Number of {self.extrema_type} is {self.num_extrema}."
+                        f" Number of {self.extrema_type} is not sufficient "
+                        f"to build frequency interpolant.")
+        if self.additional_message is not None:
+            self.message += "\n" + self.additional_message
+        super().__init__(self.message)
+
+
+class NotInRange(Exception):
```
vijayvarma392, line 31: Same, why does this exist?
Also, name is too generic...

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Because we recommend only Amplitude-based methods.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: oops! extra tab. Fixed it.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Okay, done.

gw_eccentricity/exceptions.py
md-arif-shaikh: This is to help catching exception that is solely due to insufficient extrema because generic exception can happen for many things. This could be useful when you use try except kind of statement where you can specify what kind of exception one is looking at. I have added more documentation.

gw_eccentricity/exceptions.py
md-arif-shaikh: Similarly, it helps identify the exception cause when using try and except. I have modified the name and added documentation.

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done!

gw_eccentricity/exceptions.py
vijayvarma392: I still don't get why we need this.
Isn't the idea like:
If the new flag is passed, ecc=meanano=0 is returned when appropriate.
Why does the user or a postprocessing script need to worry about handling different types of exceptions?

gw_eccentricity/eccDefinition.py
```diff
@@ -1435,7 +1544,7 @@ def derivative_of_eccentricity(self, t, n=1):
             nth order time derivative of eccentricity.
         """
         # Check that t is within tmin and tmax to avoid extrapolation
-        self.check_time_limits(t)
+        self.check_input_limits(t, self.tmin, self.tmax)
```
vijayvarma392, line 1547: Do these changes have anything to do with the current PR?

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
```diff
@@ -261,6 +261,12 @@ def load_LAL_waveform(**kwargs):
             zero_ecc_kwargs['ecc'] = 0
         zero_ecc_kwargs['include_zero_ecc'] = False   # to avoid infinite loops
         dataDict_zero_ecc = load_waveform(**zero_ecc_kwargs)
+        # To make sure that we can compute the residual amplitude/frequency, we
+        # need the zeroecc data to be longer than the ecc data so that we can
+        # intepolate the zeroecc data on the same times as the ecc data.
+        while dataDict_zero_ecc['t'][0] > dataDict["t"][0]:
+            zero_ecc_kwargs["Momega0"] *= 0.5
+            dataDict_zero_ecc = load_waveform(**zero_ecc_kwargs)
```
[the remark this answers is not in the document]
md-arif-shaikh: In our previous tests, we used an `ecc=0.1` for the `EccTD` waveform and for these using the same `Momega0` results in longer quasicircular waveforms, so we did not have any issues. But in the new test where we test the `set_failures_to_zero` with `ecc=1e-4`, I needed to add this while loop to make the quasicircular longer. So it is required for this PR. Regarding the reduction factor, yes, may be 0.9 would be better. I will check and modify.

test/test_set_failures_to_zero.py
```diff
@@ -0,0 +1,56 @@
...
+    for method in available_methods:
+        gwecc_dict = measure_eccentricity(
+            tref_in=tref_in,
+            method=method,
+            dataDict=dataDict,
+            extra_kwargs=extra_kwargs)
+        tref_out = gwecc_dict["tref_out"]
+        ecc_ref = gwecc_dict["eccentricity"]
+        meanano_ref = gwecc_dict["mean_anomaly"]
+        if method in ["Amplitude", "Frequency"]:
```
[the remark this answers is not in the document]
md-arif-shaikh: Makes sense, I will do that.

gw_eccentricity/exceptions.py
md-arif-shaikh: I think I get your point that if you avoid the extrema-related exception using the set_failures_to_zero option then probably there is no advantage in adding this custom exception. My original thinking was that irrespective of the set_failures_to_zero, the user will have a clear idea of what kinds of exception it is and even if they did not set the set_failures_to_zero, they will have the option to do whatever they want to do in such cases.

Having said that if you think it is not adding anything, I can go back to the generic exception.

gw_eccentricity/eccDefinition.py
```diff
@@ -1236,7 +1294,20 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
+
+        # If the eccentricity is too small for a method to find the extrema,
+        # and `set_failures_to_zero` is set to true, then we set the
+        # eccentricity and mean anomaly to zero and return them. In this case,
+        # the rest of the code in this function is not executed, and therefore,
+        # many variables used in diagnostic tests are never computed, making
+        # diagnostics irrelevant.
+        if any([self.probably_quasicircular_pericenter,
+                self.probably_quasicircular_apocenter]) \
+                and self.set_failures_to_zero:
```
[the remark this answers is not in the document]
md-arif-shaikh: Even if one of these is true, we will be unable to measure the eccentricity. We need both interpolants to build ecc interpolant.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: One of the comments earlier was to not copy lines and make use of functions. I found that I was repeating the checks of time and frequency inputs in the new function where we set ecc and ano to zero. So I just modified this function so that this can be now used for both time and frequency inputs.

gw_eccentricity/eccDefinition.py
```diff
@@ -237,6 +238,16 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                The code normally raises an exception if sufficient number of
+                extrema are not found. This can happen for various reasons
+                including when the eccentricity is too small for some methods
+                (like the Amplitude method) to measure. See e.g. Fig.4 of
+                arxiv.2302.11257. If `set_failures_to_zero` is set to True, we
+                assume that small eccentricity is the cause, and set the
+                returned eccentricity and mean anomaly to zero when sufficient
+                extrema are not found. USE THIS WITH CAUTION!
         """
         # Get data necessary for eccentricity measurement
```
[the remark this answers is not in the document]
md-arif-shaikh: Sure! created an issue to address later https://github.com/vijayvarma392/gw_eccentricity/issues/205

test/test_set_failures_to_zero.py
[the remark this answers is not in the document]
md-arif-shaikh: I realized that for the given `ecc=1e-4` the methods fail to detect extrema only for these two methods and therefore I test if ecc and mean ano are set to zero since that is the desired behavior.

gw_eccentricity/load_data.py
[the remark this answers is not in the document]
md-arif-shaikh: changed the factor to 0.9. Done!

gw_eccentricity/eccDefinition.py
```diff
@@ -305,7 +317,10 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         # called, these get set in that function.
         self.t_for_omega22_average = None
         self.omega22_average = None
-
+        # Approximate number of orbits derived using the phase of 22 mode
+        # assuming that a phase change 4pi occurs over an orbit.
```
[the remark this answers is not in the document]
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
```diff
@@ -1055,26 +1071,48 @@ def interp_extrema(self, extrema_type="pericenters"):
...
-                                  f"{recommended_methods}.")
+            # check if the waveform is sufficiently long
+            if self.approximate_num_orbits > 5:
+                # The waveform is sufficiently long but the extrema finding
+                # method fails to find enough number of extrema. This may
+                # happen if the eccentricity is too small and, therefore, the
+                # modulations in the amplitude/frequency is too small for the
+                # method to detect them. In such cases, the waveform can be
+                # assumed to be quasicircular.
+                probably_quasicircular = True
```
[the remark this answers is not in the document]
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
```diff
@@ -1055,26 +1071,48 @@ def interp_extrema(self, extrema_type="pericenters"):
...
         num_extrema = len(extrema)
         if num_extrema < 2:
-            recommended_methods = ["ResidualAmplitude", "AmplitudeFits"]
-            if self.method not in recommended_methods:
-                method_message = ("It's possible that the eccentricity is too "
-                                  f"low for the {self.method} method to detect"
-                                  f" the {extrema_type}. Try one of "
-                                  f"{recommended_methods}.")
+            # check if the waveform is sufficiently long
+            if self.approximate_num_orbits > 5:
```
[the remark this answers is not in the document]
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
```diff
@@ -1055,26 +1071,48 @@ def interp_extrema(self, extrema_type="pericenters"):
             return self.get_interp(self.t[extrema],
                                    self.omega22[extrema])
         else:
-            raise Exception(
-                f"Sufficient number of {extrema_type} are not found."
-                " Can not create an interpolant.")
+            raise InsufficientExtrema(extrema_type, len(extrema))
 
     def check_num_extrema(self, extrema, extrema_type="extrema"):
         """Check number of extrema."""
```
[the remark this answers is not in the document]
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
```diff
@@ -1222,10 +1260,30 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
         # Get the pericenters and apocenters
         pericenters = self.find_extrema("pericenters")
         original_pericenters = pericenters.copy()
-        self.check_num_extrema(pericenters, "pericenters")
+        # Check if there are a sufficient number of extrema. In cases where the
+        # waveform is long enough but the method fails to detect any extrema,
+        # it might be that the eccentricity is too small for the current method
+        # to detect it. See Fig.4 in arxiv.2302.11257. In such cases, we assume
+        # that the waveform is probably quasicircular.
+        self.probably_quasicircular_pericenter = self.check_num_extrema(
```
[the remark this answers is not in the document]
md-arif-shaikh: These are used in a function outside this, so it is convenient in this case to have it available.

gw_eccentricity/exceptions.py
vijayvarma392: Yeah, I think get rid of it. Might lead to confusion while solving a problem that's not a real problem. Same for the other thing in the exceptions.py file?

test/test_set_failures_to_zero.py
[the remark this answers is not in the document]
vijayvarma392: Ok, is there a test you can come up with for the other methods as well? Like does ecc=1e-6 work for that model? If you do come up with such a case, keep 1e-4 for the amp/freq method because that's most relevant for them. If not, just explain in the test why you are only testing on a subset of method.

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
vijayvarma392: Can you point me to where else this is being used?

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
vijayvarma392: Ok, I think your comment tripped me up. It's not that the diagnostics are not "irrelevant", it's more like we don't need the additional checks as we will simply return zero for ecc and mean ano?

"many variables used in diagnostic tests are never computed, making diagnostics irrelevant"
suggests that some of the diagnostic plots may not work if this option is used. Is that what you mean?

Either way, this comment needs more clarity.

gw_eccentricity/eccDefinition.py
vijayvarma392: Ok, fine for now, but that's a perfect example of: make an issue and deal with later.

gw_eccentricity/exceptions.py
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: It is in the check_input_limits. I have however removed these two from self and instead added a variable `eccentricity_and_mean_anomaly_have_been_set_to_zero` so that depending on whether it's True or False, we can display relevant information.

test/test_set_failures_to_zero.py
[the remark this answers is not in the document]
md-arif-shaikh: I used `IMRPhenomT` to generate the `dataDict`. Since it's a quasicircular waveform, none of the methods will be able to find any extream and this can be used to test if the ecc and mean ano are set to zero.

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: Ok, rephrased it.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Ok, will do that for future PRs.

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
vijayvarma392: Yes, that's better. Although question: is there ever a time when this variable has been set to True, but you still haven't actually set them to zero. Or the other way around. To avoid that confusion, would it be better to call this `setting_ecc_to_zero`? We don't have to explicitly say the same for meanano in the variable name.

test/test_set_failures_to_zero.py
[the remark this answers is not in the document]
vijayvarma392: Yes, go for it. But do that only for the residual type methods, because we want to test Amp/Freq at 1e-4.

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: Yes, I will think about it and try to implement a fix in a separate PR.

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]

test/test_set_failures_to_zero.py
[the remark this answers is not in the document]
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
```diff
@@ -237,6 +237,16 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
                 eccDefinitionUsingFrequencyFits.get_default_kwargs_for_fits_methods
                 for allowed keys.
+
+            set_failures_to_zero : bool, default=False
+                The code normally raises an exception if sufficient number of
+                extrema are not found. This can happen for various reasons
+                including when the eccentricity is too small for some methods
+                (like the Amplitude method) to measure. See e.g. Fig.4 of
+                arxiv.2302.11257. If `set_failures_to_zero` is set to True, we
+                assume that small eccentricity is the cause, and set the
```
vijayvarma392, line 247: Actually we should be more clear:
assume that small eccentricity is the cause as long as the waveform has at least 5 orbits
Question: What is the actual rule. We take a waveform, truncate 2 orbits at the end, and then look for 5 orbits in this test? So, is it min 7 orbits? Whatever it is, give the total number here (7?), and in the code where you have stuff like 5 orbits, explain what that means via a comment.

gw_eccentricity/eccDefinition.py
```diff
@@ -1060,21 +1071,85 @@ def interp_extrema(self, extrema_type="pericenters"):
...
-        """Check number of extrema."""
+        """Check number of extrema.
+
+        Check the number of extrema to determine if there are enough for
+        building the interpolants through the pericenters and apocenters. In
+        cases where the number of extrema is insufficient, i.e., less than 2,
+        we further verify if the provided waveform is long enough to have a
+        sufficient number of extrema.
+
+        If the waveform is long enough, but the chosen method fails to detect
```
vijayvarma392, line 1082: Here is where you should explain the 5 orbits thing, and if that's really 7 orbits -- but explain this using the variables like num_orbits_to_remove_before_merger, and their default values. Basically, help out the user in figuring out how long the input waveform needs to be to use this option.

gw_eccentricity/eccDefinition.py
```diff
@@ -1222,10 +1297,30 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
+            self.fref_in = np.atleast_1d(fref_in)
         # Get the pericenters and apocenters
         pericenters = self.find_extrema("pericenters")
         original_pericenters = pericenters.copy()
-        self.check_num_extrema(pericenters, "pericenters")
+        # Check if there are a sufficient number of extrema. In cases where the
+        # waveform is long enough but the method fails to detect any extrema,
+        # it might be that the eccentricity is too small for the current method
+        # to detect it. See Fig.4 in arxiv.2302.11257. In such cases, we assume
+        # that the waveform is probably quasicircular.
```
vijayvarma392, line 1321: The last sentence might need to be updated?

gw_eccentricity/eccDefinition.py
```diff
@@ -1350,26 +1442,85 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
-        return_dict = {"eccentricity": self.eccentricity,
-                       "mean_anomaly": self.mean_anomaly}
-        if fref_in is not None:
-            return_dict.update({"fref_out": self.fref_out})
+        # return measured eccentricity, mean anomaly and reference time or
+        # frequency where these are measured.
+        return self.make_return_dict_for_eccentricity_and_mean_anomaly()
+
+    def set_eccentricity_and_mean_anomaly_to_zero(self):
+        """Set eccentricity and mean_anomaly to zero."""
```
vijayvarma392, line 1455: I'm thinking the following:
What if we just return an array of zeros, the same size as tref_in/fref_in and be done with it?
If the user asks for a tref_in/fref_in that is outside t, it's not a big deal because zero ecc remains zero ecc for all times. That way, this function is very simple, and you don't have to make assumptions about instantaneous f_22 being a decent option.

gw_eccentricity/eccDefinition.py
```diff
@@ -1350,26 +1442,85 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
+        then the returned eccentricity and mean anomaly are also scalars. To do
+        this, we use the information about tref_in/fref_in that is provided by
+        the user. At the top of the measure_ecc function, we set ref_ndim to
+        identify whether the original input was scalar or array-like and use
+        that here.
+        """
+        # If the original input was scalar, convert the measured eccentricity,
+        # mean anomaly, etc., to scalar.
+        if self.ref_ndim == 0:
+            self.eccentricity = self.eccentricity[0]
```
vijayvarma392, line 1505: Shouldn't we double check that the size of the original array was 1 before converting to scalar?

test/test_set_failures_to_zero.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_set_failures_to_zero():
+    """Test handling of failures due to insufficient extrema.
+
+    In certain situations, the waveform may have zero eccentricity or a very
+    small eccentricity, making it difficult for the given method to identify
+    any extrema. In cases where such a situation occurs, and if the user has
+    set 'set_failures_to_zero' to `True` in the 'extra_kwargs' parameter, both
+    the eccentricity and mean anomaly will be set to zero.
```
vijayvarma392, line 14: This doesn't say what this is testing: it's testing that the interface works with set_failures_to_zero for waveforms with small or zero ecc.

test/test_set_failures_to_zero.py
```diff
@@ -0,0 +1,69 @@
...
+    lal_kwargs_qc = lal_kwargs_ecc.copy()
+    lal_kwargs_qc.update({"approximant": "IMRPhenomT",
+                          "ecc": 0.0})
+    # Create the dataDict for Amplitude and Frequency method
+    dataDict_ecc = load_data.load_waveform(**lal_kwargs_ecc)
+    # Create the dataDict for Residual and Fits method
+    dataDict_qc = load_data.load_waveform(**lal_kwargs_qc)
+    available_methods = gw_eccentricity.get_available_methods()
+    tref_in = -8000
+    fref_in = 0.005
```
vijayvarma392, line 40: Why just a scalar? We should also test that we get an array of zeros when an array of times/freqs is passed.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done!

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Okay, done!

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

test/test_set_failures_to_zero.py
md-arif-shaikh: Done

test/test_set_failures_to_zero.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
```diff
@@ -243,10 +243,21 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                conditions are met:
+
+                1. `set_failures_to_zero` is set to `True`.
+                2. The length of the waveform is required to be at least (5 +
+                `num_obrits_to_exclude_before_merger`) orbits long. By default,
+                `num_obrits_to_exclude_before_merger` is set to 2, meaning that
+                2 orbits are removed from the waveform before it is used by the
+                extrema finding routine. Consequently, in the default
+                configuration, the original waveform in the input `dataDict`
+                must have a minimum length of 7 orbits.
```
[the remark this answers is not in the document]

gw_eccentricity/eccDefinition.py
```diff
@@ -243,10 +243,21 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+
+                we assume that small eccentricity is the cause, and set the
```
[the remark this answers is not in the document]

vijayvarma392 (approved):

PR #204, opened 2023-08-31
add method to load newer sxscatalog files

md-arif-shaikh:
`load_sxs_catalogformat` is modified to allow loading sxs waveform files in the new format.


gw_eccentricity/load_data.py
```diff
@@ -827,22 +834,24 @@ def load_sxs_catalogformat(**kwargs):
...
-        modes_dict.update({(ell, m): hlm_interp})
+    # Check file format
+    if "Strain_N" in kwargs["filepath"]:
+        # find the extrapolation order from file name and check if
+        # it is consistent with the value provided via kwargs.
+        # If they are not consistent, raise warnings and ignore
+        # the value provided in the kwargs.
+        base_name = os.path.basename(kwargs["filepath"])
+        extrap_order_from_filename = int(
+            re.findall(r"\d", base_name)[0])
```
vijayvarma392, line 845: Just to be safer, search for the exact format. Something like
```
matches = re.findall("StrainN(\d).h5")
if len(matches) != 1:
    raise Exception...
extrap_order_from_filename = int(matches[0])
```

gw_eccentricity/load_data.py
```diff
@@ -827,22 +834,24 @@ def load_sxs_catalogformat(**kwargs):
...
+        if kwargs["extrap_order"] != extrap_order_from_filename:
+            warnings.warn("Exrapolation order in `kwargs` is "
+                          f"{kwargs['extrap_order']} but waveform file seems "
+                          "to have extrapolation order "
+                          f"{extrap_order_from_filename}. "
+                          "Ignoring the `kwargs` value.")
```
vijayvarma392, line 851: Instead of this complicated stuff...don't allow extrap_order in kwargs for the new format. Error is more appropriate here.

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
md-arif-shaikh: It becomes trickier. Currently, the idea is that we use the same origin="SXSCatalog" in the load_waveform and depending on the file format, we use either the load_old/load_new. In both cases, they use the same set of default kwargs. But if we want to change the default kwargs, i.e., allow extrap_order in old but not in the new, then we have to change the origin to "SXSCatalog_old" and "SXSCatalog_new" since the default kwargs depend on the origin. Also, in that case, we don't need the common load function but instead just use the separate ones. If we go that route then the user needs to figure out which format it is beforehand and the effort to make a single sxs load function is needed anymore.

gw_eccentricity/load_data.py
md-arif-shaikh: Done

gw_eccentricity/load_data.py
```diff
@@ -835,7 +835,8 @@ def load_sxs_catalogformat(**kwargs):
                 "used to eavaluate the zero ecc waveform.")
 
     # Check file format
-    if "Strain_N" in kwargs["filepath"]:
+    matches = re.findall(r"Strain_N\d.h5", kwargs["filepath"])
+    if len(matches) == 1:
```
vijayvarma392, line 839: This way you don't catch cases where len(matches) != 1, right? You need an Exception. Also, hopefully you are testing this code?

gw_eccentricity/load_data.py
vijayvarma392: Ok, but this should still be an exception.
 
But more importantly, you are telling me the code will only work for extrap_order=3 if the user gets it right both in the kwargs and in the filename? If the user does the natural thing: give the file path and because it has the extrap order in it, assume the code is smart enough to figure it out and therefore not specify extrap_order in the kwargs. Then, because it will default to 2 in the kwargs, things will be inconsitent and the user gets a warning/exception? Not nice.

How about the much simpler option of:
Say in the docs for the kwargs that extrap order will be ignored for the new format. Then, just don't check for it.

vijayvarma392:
BTW, does the example notebook need to be updated as well, to show how to use the new format files?

gw_eccentricity/load_data.py

gw_eccentricity/load_data.py

md-arif-shaikh:
> BTW, does the example notebook need to be updated as well, to show how to use the new format files?

Done

gw_eccentricity/load_data.py
```diff
@@ -712,9 +724,12 @@ def load_lvcnr_waveform(**kwargs):
...
     """Load modes from sxs waveform files in sxs catalog format.
 
-    This function is intended for loading waveform modes from files in
-    the sxs catalog format (see
+    This function is intended for loading waveform modes from files in the sxs
+    catalog in the new format, i.e., from 2023 onward (see
     https://data.black-holes.org/waveforms/documentation.html).
+
+    For loading sxs catalog waveforms in old format see
+    `load_sxs_catalogformat_old`.
```
vijayvarma392, line 732: Be more helpful: say old means rhOhverM..., new means Strain..

gw_eccentricity/load_data.py
```diff
@@ -901,9 +1138,17 @@ def get_params_dict_from_sxs_metadata(metadata_path):
             m2 = float(line.split("=")[-1].strip())
     # numerical noise can make m1 slightly lesser than m2. Catch this
     # whenver it happens
-    if m1 < m2:
-        raise Exception(f"SXS metadata gives m1 = {m1} < m2 = {m2}")
-    params_dict = {"q": m1/m2,
+    q = m1/m2
+    if q < 1:
+        # if q > 0.99, treat it as 1.
+        if q > 0.99:
```
vijayvarma392, line 1144: As I was saying, 0.99 is not high enough. What's the actual number for q you are getting? Can we make it 1-1e-5?

gw_eccentricity/load_data.py
```diff
@@ -1030,6 +1275,100 @@ def reomve_junk_from_nr_data(t, modes_dict, num_orbits_to_remove_as_junk):
...
+    # Check if it is true, i.e., z-coordinate remains very small
+    if any(np.abs(separion_vec[:, 2]) > 1e-5):
+        raise Exception(
+            "The system seems to be precessing. Cannot get the orbital "
+            "phase by assuming the motion is restricted in the x-y plane.")
+    # Get the orbital phase
+    phase_orb = np.unwrap(np.arctan2(separion_vec[:, 1], separion_vec[:, 0]))
+    # Interpolate phase
+    time_interp = np.arange(min(time), max(time), deltaTOverM)
+    phase_interp = InterpolatedUnivariateSpline(time, phase_orb)(time_interp)
```
vijayvarma392, line 1362: I don't get it, why do you need to interpolate here?

gw_eccentricity/load_data.py
```diff
@@ -1030,6 +1275,100 @@ def reomve_junk_from_nr_data(t, modes_dict, num_orbits_to_remove_as_junk):
...
+    # Find the duration of first num_orbits assuming that the orbital phase
+    # changes by 2pi over one orbit
+    idx_at_num_obits_from_start = np.argmin(
+        np.abs(phase_interp - (phase_interp[0] + num_orbits * 2 * np.pi)))
+    num_obits_duration = (time_interp[idx_at_num_obits_from_start]
+                          - time_interp[0])
+    return num_obits_duration
```
vijayvarma392, line 1369: Have you tested that this solves the problem you are trying to solve, let's say for the Ecc3dSur dataset?

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
```diff
@@ -161,7 +165,10 @@ def load_waveform(origin="LAL", **kwargs):
         - "LAL": Compute waveform by a call to the LAL-library.
             (see https://lscsoft.docs.ligo.org/lalsuite/lalsimulation/)
         - "SXSCatalog": Import waveform by reading a file in the SXS catalog
-            format.
+            in the new format (from 2023 onward).
```
vijayvarma392, line 168: Be more helpful, say what new and old means in terms of the h5 filenames.

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
md-arif-shaikh: Done

gw_eccentricity/load_data.py
md-arif-shaikh: Because the NR times are not uniform. Maybe it would not make a huge difference but don't know how large the deltaT in NR files could be, so thought it would be safe to use interpolation.

gw_eccentricity/load_data.py
md-arif-shaikh: Done

gw_eccentricity/load_data.py
vijayvarma392: Nah, uniform times doesn't matter here. You just want the orbital time scale, and the time step will never be so large that the binaries jump a huge fraction of an orbit.

md-arif-shaikh: Okay, removed that.

gw_eccentricity/load_data.py
md-arif-shaikh: I have not tried the full set yet. I just tried the one I was having a problem with. Here is one plot. I clearly show that Horizons data gives a much better result
image

gw_eccentricity/load_data.py
md-arif-shaikh: I think it need not be 0.99 and in fact, the difference is very small close to what you are suggesting. So, I will go through the Ecc1d runs since these have q=1 and find the typical value of 1-q

vijayvarma392: And the one that is 1e-4, is q=1-1e-4, and not q=1+1e-4, right?
If yes, let's go with 5e-4 for now, and we can adjust later if needed.

gw_eccentricity/load_data.py
vijayvarma392: Is that for two orbits? Visually, it still looks like less than two. 
I might be because:
The time array in Horizons.h5 is the time coordinate of the simulation, so it starts at t=0.
The time array in the waveform file is related to the retarded, shifted back to the origin, which means it starts at some t<0, like order -100M or so. However, the t=0 point in both of these files can be approximately associated with each other.

What if you do: get DeltaT from Horizons.h5, then truncate all times up to t_waveform = DeltaT, not t_waveform = t_waveform[0] + DeltaT. Does that make it better?

gw_eccentricity/load_data.py
md-arif-shaikh: Okay, done!

gw_eccentricity/load_data.py
md-arif-shaikh: Okay, I will check this.

vijayvarma392:
Also, talk about the Horizons.h5 file in the notebook:
image

Think of it from the perspective of a new user -- you need to tell them which whiles are good to include.

And make sure the docs printed in the notebook are up to date:
image
This should already be explaining what new/old means.

gw_eccentricity/load_data.py
md-arif-shaikh: I shared some plots here. The first thing we see is that the duration of the waveform is random as it is obtained using noisy data. Comparing the duration from Horizons.h5, from $t=0$, using just $t >$ duration seems to give a better result than using $t > t[0]$ + duration. What do you think?

image
image
image
image

md-arif-shaikh:
> Also, talk about the Horizons.h5 file in the notebook: image
> 
> Think of it from the perspective of a new user -- you need to tell them which whiles are good to include.
> 
> And make sure the docs printed in the notebook are up to date: image This should already be explaining what new/old means.

I will run and push the updated notebook once the orbit thing is fixed.

gw_eccentricity/load_data.py
vijayvarma392: Yeah, this is for the reason I mentioned above. Let's go with the pink line. But, you should test this more to be safe, like for the EccHybTest dataset.

md-arif-shaikh: The pink line looks good for the `EccHybTest` cases also
image
image
image

vijayvarma392: sounds good

vijayvarma392 (approved):

Issue #206, opened 2023-10-17
Handing failures due to eccentricity decay


md-arif-shaikh:
PR #203 addresses #84 when no sufficient extrema are not found and we can not build interpolants through the pericenters and apocenters. However, there would be cases where we can find enough extrema in the initial part of the waveform and therefore build the interpolants but due to decay of eccentricity the last pericenters/apocenters may be far from the merger. In such cases, if `tref_in` is later than the last `pericenters/apocenters`, eccentricity measurement will fail. We need to handle such failures also.

PR #208, opened 2023-10-24
set ecc to zero for failures t/f > tmax/fmax

md-arif-shaikh:
Addressing #206. For methods like `Amplitude` or `Frequency`, it could be challenging to detect the small oscillations in the later part of the inspiral and therefore the maximum allowed time `tmax` or frequency `fmax` could be much smaller than the time/frequency near merger. Therefore, for time/frequency after tmax/fmax, eccentricity can not be measured. If `set_failures_to_zero` is True, we set the eccentricity/mean anomaly to zero for such cases at these times/frequencies.

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -8,6 +8,7 @@ https://github.com/vijayvarma392/gw_eccentricity/wiki/Adding-new-eccentricity-de
 
 import numpy as np
 import matplotlib.pyplot as plt
+import warnings
 from .utils import peak_time_via_quadratic_fit, check_kwargs_and_set_defaults
 from .utils import amplitude_using_all_modes
 from .utils import time_deriv_4thOrder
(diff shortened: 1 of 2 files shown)


gw_eccentricity/eccDefinition.py
```diff
@@ -258,6 +263,34 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                'Frequency.'  In such cases, `tmax` and `fref_max` may become
+                significantly smaller than the time or frequency near the
+                merger.
+
+                Consequently, eccentricity measurement becomes unsuccessful
+                for any time or frequency in the late stages of the inspiral
+                (as illustrated in Figure 4 of arxiv.2302.11257). If the
+                `set_failures_to_zero` parameter is set to True, the code
+                will set both eccentricity and mean anomaly to zero in these
+                instances.
```
vijayvarma392, line 292: Do we need all this?
Can't we just do: If no extrema are found but the waveform is long enough, return an array of zeros the same size as tref_in/fref_in without even checking where they fall w.r.t. tmin/tmax. Don't we already do this? I don't think I see a down side, because we have "determined" the waveform to be of zero ecc, and zero ecc remains zero ecc for all tref_in/fref_in.

md-arif-shaikh: I think you may have missed the text under issue #206. This is different from #84 in the sense that in this case, we have sufficient extrema to build interpolants and measure eccentricity. But since the initial eccentricity is small we can find the extrema only till some tmax and it will fail if we want to measure ecc anywhere > tmax. In the previous case there was tmin/tmax to begin with because we did not have enough extrema.

md-arif-shaikh: For example, with the following params
```python
lal_kwargs = {"approximant": "EccentricTD",
                  "q": 1.0,
                  "chi1": [0.0, 0.0, 0.0],
                  "chi2": [0.0, 0.0, 0.0],
                  "Momega0": 0.01,
                  "ecc": 0.01,
                  "mean_ano": 0,
                  "include_zero_ecc": True}
```
The `tmax` is about `-3000M`, so it will fail for `tref_in=-2500M`. In this PR, we can return `ecc=0` for such cases if `set_failures_to_zero` is True.

md-arif-shaikh: What you mentioned in the comment is exactly what we are doing when no extrema are found. We don't check tmin/tmax. We just return zeroes for whatever tref_in is.

vijayvarma392: Makes sense thanks!

vijayvarma392:
Kill this PR?

md-arif-shaikh:
Closing this PR as it has gone stale and not planning to work on it anytime soon!

PR #210, opened 2024-01-11
remove non-precessing condition

md-arif-shaikh:
Requiring the system to be non-precessing does not work for loading SXS waveform modes for processing cases. We remove this condition and compute the phase from the projected motion on the x-y plane.

### DIFF
--- a/gw_eccentricity/load_data.py
+++ b/gw_eccentricity/load_data.py
@@ -1369,12 +1369,15 @@ def get_num_orbits_duration_from_horizon_data(horizon_filepath, num_orbits):
     # We will use the x-y coordinates, i.e., the motion projected
     # onto the x-y plane to compute the phase and then the number of
     # orbits using that phase.
+    # NOTE: This assumption breaks down when the orbital plane itself contains
+    # the z-axis, which can happen if there is extreme precession, or in the
+    # case of EMRIs.
     # Get the orbital phase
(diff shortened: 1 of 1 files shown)


gw_eccentricity/load_data.py
```diff
@@ -1366,13 +1366,9 @@ def get_num_orbits_duration_from_horizon_data(horizon_filepath, num_orbits):
...
-    # We will assume the system to be non-precessing so that the motion
-    # is restricted in the x-y plane and the z-coordinate is effectively zero.
-    # Check if it is true, i.e., z-coordinate remains very small
-    if any(np.abs(separion_vec[:, 2]) > 1e-5):
-        raise Exception(
-            "The system seems to be precessing. Cannot get the orbital "
-            "phase by assuming the motion is restricted in the x-y plane.")
+    # We will use the x-y coordinates, i.e., the motion projected
+    # onto the x-y plane to compute the phase and then the number of
+    # orbits using that phase.
```
vijayvarma392, line 1371: Call it phase_orb_projected so that it is not accidentally used for something else.

vijayvarma392: Maybe add a comment saying that this assumption breaks down when the orbital plane itself contains the z-axis, which can happen if there is extreme precession, or in the case of EMRIs.

md-arif-shaikh: Done

vijayvarma392 (approved):

Issue #211, opened 2024-03-18
warn if phase is not increasing


md-arif-shaikh:

md-arif-shaikh:
closed with #212

PR #212, opened 2024-03-19
check phase22 sign convention

md-arif-shaikh:
Phase22 should be monotonically increasing. Check whether the phase22 at the merger is greater than that at the start or not. #211

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -588,13 +588,18 @@ class eccDefinition:
         # than the phase22 at the start of the waveform
         if phase22_merger < phase22[0]:
             raise Exception(
-                f"Phase22 = {phase22_merger} at the merger is < "
+                f"phase22 = {phase22_merger} at the merger is < "
                 f"phase22 = {phase22[0]} at the start. The "
                 "phaselm should be related to hlm as "
                 "hlm = amplm * exp(- i phaselm) ensuring that "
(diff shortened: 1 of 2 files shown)


gw_eccentricity/eccDefinition.py
```diff
@@ -584,6 +584,17 @@ def process_data_dict(self,
         amp22_merger = newDataDict["amplm"][(2, 2)][merger_idx]
         phase22 = newDataDict["phaselm"][(2, 2)]
         phase22_merger = phase22[merger_idx]
+        # check if phase22 is increasing. phase22 at merger should be greater
+        # than the phase22 at the start of the waveform
+        if phase22_merger < phase22[0]:
+            raise Exception(
+                f"Phase22 = {phase22_merger} at the merger is < "
```
vijayvarma392, line 591: Phase22 --> phase22, to be consistent with our own documentation:
https://github.com/vijayvarma392/gw_eccentricity/blob/b96e69dd373dca54fe18ed71676aabda863ff753/gw_eccentricity/eccDefinition.py#L38

vijayvarma392:
@md-arif-shaikh, we should have a similar check that omega22 is positive. Checking that at tmerger is reasonable.

gw_eccentricity/eccDefinition.py

md-arif-shaikh:
Both are done @vijayvarma392

vijayvarma392 (approved):

Issue #213, opened 2024-04-29
Actions broken


vijayvarma392:
You should be able to fix it like this https://github.com/vijayvarma392/surfinBH/commit/fbff946117cc42d7dc55126b94ce710da174095b

vijayvarma392:
any luck with this?

md-arif-shaikh:
Failing due to some issue while building `spinsfast` package
![image](https://github.com/vijayvarma392/gw_eccentricity/assets/44096766/f23d151c-cf1d-4089-9ecf-373f6922b86b)

md-arif-shaikh:
maybe of changes made in spinsfast last week?
![image](https://github.com/vijayvarma392/gw_eccentricity/assets/44096766/a2b88861-77c9-4192-844f-9bb03ac58a3f)

md-arif-shaikh:
can we temporarily remove `sxs` from dependency? The tests do not use `sxs`. `spinsfast` is required by `sxs`, so removing `sxs` may be a temporary fix. @vijayvarma392

vijayvarma392:
The real error might be something else:
<img width="1265" alt="image" src="https://github.com/vijayvarma392/gw_eccentricity/assets/9355891/50859a69-038c-4cdd-8855-9be88c48b335">

Is sxs an actual dependency for the package? If so, we shouldn't just remove it. 

Looks like only the macos test was run though: https://github.com/vijayvarma392/gw_eccentricity/actions/runs/8890734138/job/25533440833?pr=214. Can you try disabling macos test so that the ubuntu test is forced to run?

vijayvarma392:
Also, maybe worth trying newer python versions: https://github.com/moble/spinsfast/blob/main/.github/workflows/build.yml#L18

md-arif-shaikh:
Should this be closed now @vijayvarma392?

vijayvarma392:
Yes, and feel free to take executive decisions about which issues can be closed :)

PR #214, opened 2024-04-30
fix action #213

md-arif-shaikh:


md-arif-shaikh:
with python=3.10 ubutu succeeded but macos failed. Lalsuite did not work for 3.12. @vijayvarma392

md-arif-shaikh:
I found a fix. For macos, the trick was to use `brew` to install `fftw` first and then install `spinsfast` with `CFLAGS=" -L/opt/homebrew/lib " python -m pip install spinsfast` as suggested in `spinsfast` repo [here](https://github.com/moble/spinsfast?tab=readme-ov-file#pip)  @vijayvarma392

vijayvarma392 (approved):

PR #215, opened 2024-06-06
replace omega22 -> omega_gw

md-arif-shaikh:
To accommodate the precessing case, we need to change `omega22` to a generic name `omega_gw`. This is because unlike the nonprecessing case, 22 mode is neither dominant nor the most suitable quantity for precessing systems to measure eccentricity from. For example, in the precessing case, a combination of omega from (2, 2) and (2, -2) mode in the co-precessing frame may provide a more suitable omega to use for measuring eccentricity.


vijayvarma392:
Is this ready for review? I don't always get notifications unless you ask for a review.

vijayvarma392:
Cool. Tip: You can set such PRs to draft to avoid confusion:
image

md-arif-shaikh:
Thanks! I forgot about this feature.

vijayvarma392:
Should this be 
omega_gw_average defaults to <omega22>?
image

vijayvarma392:
How about setting omega_gw=omega_22 in the ylabel here?
image

And then, when you do something else for precession, change the label accordingly.

vijayvarma392:
as described in Eq.(4) of arXiv:2302.11257 (e_omega_gw is called e_omega_22 in the paper).
image

vijayvarma392:
Maybe worth taking one paragraph to make things extra clear right after this:
image

Like: Throughout this documentation, we will refer to omega_gw and amp_gw. For nonprecessing,..... For precessing, ....

Then, in the other places, you don't have to give a full explanation every time (occasional quick reminders are good).

gw_eccentricity/eccDefinition.py
```diff
@@ -271,10 +289,16 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
                             f"Time steps are {self.t_diff}")
-        # get amplitude, phase and omega of 22 mode
-        self.amp22 = self.dataDict["amplm"][(2, 2)]
-        self.phase22 = self.dataDict["phaselm"][(2, 2)]
-        self.omega22 = self.dataDict["omegalm"][(2, 2)]
+        # get amplitude, phase, omega to be used for eccentricity measurement.
+        # For precessing systems, even in the coprecessing frame, the (2, 2)
+        # mode quantities show some oscillations due to the precession
+        # effect. To reduce these oscillations further, we use a combination of
+        # (2, 2) and (2, -2) mode in the coprecessing frame to define new
```
vijayvarma392, line 296: Instead of saying "a combination", which is vague, just point to the equation in the surrogate paper.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py

md-arif-shaikh:
> Maybe worth taking one paragraph to make things extra clear right after this: image
> 
> Like: Throughout this documentation, we will refer to omega_gw and amp_gw. For nonprecessing,..... For precessing, ....
> 
> Then, in the other places, you don't have to give a full explanation every time (occasional quick reminders are good).

Done

md-arif-shaikh:
> as described in Eq.(4) of arXiv:2302.11257 (e_omega_gw is called e_omega_22 in the paper). image

Done

md-arif-shaikh:
> How about setting omega_gw=omega_22 in the ylabel here? image
> 
> And then, when you do something else for precession, change the label accordingly.

Done

md-arif-shaikh:
> Should this be omega_gw_average defaults to ? image

Did not get this. The current line implies that by default `omega_gw_average` means the orbit averged `omega_gw` which is denoted as $\langle \omega_{\text{gw}} \rangle$

vijayvarma392:
> > Should this be omega_gw_average defaults to ? image
> 
> Did not get this. The current line implies that by default `omega_gw_average` means the orbit averged `omega_gw` which is denoted as ⟨ωgw⟩

The math part got cut out, I meant:
Should this be 
omega_gw_average defaults to `<omega22>`?

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    Throughout this documentation, we will refer to omega_gw and amp_gw. For
+    nonprecessing systems, these quantities reduce to the corresponding values
+    obtained using the (2, 2) mode, i. e., for nonprecessing systems,
+
+    amp_gw = amp22, phase_gw = phase22 and omega_gw = omega22
+
+    On the other hand, for precessing systems, we use Eq.(48) and (49) of
+    arXiv:1701.00550 to define amp_gw and phase_gw. amp_gw (phase_gw) is
+    defined using a symmetric (antisymmetric) combination of
+    amplitude (phase) of (2, 2) and (2, -2) mode.
```
vijayvarma392, line 85: Use [] instead of () to avoid mixing it up with (2,2) and (2,-2).
Say coprecessing.

vijayvarma392: If you haven't already, please do a full read thru of the docs to see if things are clear.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
```diff
@@ -154,15 +155,27 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                 https://github.com/vijayvarma392/gw_eccentricity/wiki/NR-investigation-to-set-default-number-of-orbits-to-exclude-before-merger
                 Default: 2.
 
+        precessing: bool, default=False
+            Whether the system is precessing or not. For precessing systems,
+            the `dataDict` should contain modes computed in the intertial frame
+            which are rotated internally to obtain the same in the coprecessing
+            frame. Finally the modes in the coprecessing frame are used for
+            measuring eccentricity.
```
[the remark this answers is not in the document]
md-arif-shaikh: Does the following rephrasing work?

image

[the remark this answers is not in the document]
vijayvarma392: Hmm, I thought you were not doing the rotation in the package for now? Could you point me to where this is done? Also, "For precessing....should contain modes contained in the inertial frame" suggests you want something else for nonprecessing.

md-arif-shaikh:
> > > Should this be omega_gw_average defaults to ? image
> > 
> > 
> > Did not get this. The current line implies that by default `omega_gw_average` means the orbit averged `omega_gw` which is denoted as ⟨ωgw⟩
> 
> The math part got cut out, I meant: Should this be omega_gw_average defaults to `<omega22>`?

Okay, got it. Done

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: Yes, it is not done right now. Right now `precessing=True` should not work. So technically the above documentation is not correct right now. But this is what we intend to do in future when precessing cases will be supported. Should I add a note here that only non-precessing case is supported right now?

The second para says for nonprecessing system there is no distinction between inertial and coprecessing.

Basically what I want to say is that if `precessing=True`, I want to ensure that the modes are in the intertial frame otherwise, I don't know whether I need to apply rotation or not.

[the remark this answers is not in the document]
vijayvarma392: I don't think the docs should say things that aren't done yet, just say things like: For precessing, the input waveform should already be in the coprecessing frame..stuff like that. You can update it when the code actually handles inertial frame precessing waveforms.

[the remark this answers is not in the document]
vijayvarma392: What do you mean `precessing=True` should not work? Why can't we just pass it a copr waveform and have it work for that?

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: > Use [] instead of () to avoid mixing it up with (2,2) and (2,-2). Say coprecessing.

Done

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: > What do you mean `precessing=True` should not work? Why can't we just pass it a copr waveform and have it work for that?

It should in principle work now. But not tested as explained above.

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
-    mode. We currently only allow time-domain, nonprecessing waveforms. We
-    evaluate omega22(t) at pericenter times, t_pericenters, and build a
-    spline interpolant omega22_pericenters(t) using those data
-    points. Similarly, we build omega22_apocenters(t) using omega22(t) at
-    the apocenter times, t_apocenters.
+    Eccentricity is measured using the GW frequency omega_gw(t) =
+    d(phase_gw)/dt, where phase_gw(t) is the phase of the (2, 2) waveform mode
+    for nonprecessing systems. For precessing systems, phase_gw is obtained
+    using an antisymmetric combination of (2, 2) and (2, -2) mode phases in the
+    coprecessing frame.
```
vijayvarma392, line 75: Get rid of " =
    d(phase_gw)/dt, where phase_gw(t) is the phase of the (2, 2) waveform mode
    for nonprecessing systems. For precessing systems, phase_gw is obtained
    using an antisymmetric combination of (2, 2) and (2, -2) mode phases in the
    coprecessing frame."

You repeat that bit below anyway. Just merge the next paragraph with this after deleting that bit.

That is:
Eccentricity is measured using the GW frequency omega_gw(t) = d(phase_gw)/dt. Throughout this...

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
 
-    Using omega22_pericenters(t) and omega22_apocenters(t), we first
-    compute e_omega22(t), as described in Eq.(4) of arXiv:2302.11257. We
-    then use e_omega22(t) to compute the eccentricity egw(t) using Eq.(8)
-    of arXiv:2302.11257. Mean anomaly is defined using t_pericenters, as
-    described in Eq.(10) of arXiv:2302.11257.
+    Throughout this documentation, we will refer to omega_gw and amp_gw. For
```
vijayvarma392, line 76: we will refer to phase_gw, omega_gw and amp_gw

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    nonprecessing systems, these quantities reduce to the corresponding values
```
vijayvarma392, line 77: reduce to --> are simply

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    obtained using the (2, 2) mode, i. e., for nonprecessing systems,
```
vijayvarma392, line 78: obtained using the --> of the
You already said "nonprecessing systems", get rid of that.

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    arXiv:1701.00550 to define amp_gw and phase_gw. amp_gw [phase_gw] is
+    defined using a symmetric [antisymmetric] combination of
+    amplitude [phase] of (2, 2) and (2, -2) mode.
+
+    amp_gw = (1/2) * (amp(2, 2) + amp(2, -2))
+    phase_gw = (1/2) * (phase(2, 2) - phase(2, -2))
+    omega_gw = d(phase_gw)/dt
+
+    These quantities reduce to the corresponding (2, 2) mode data when the
+    system is nonprecessing. See `eccDefinition.get_amp_phase_omega_gw` for
```
vijayvarma392, line 92: "These quantities reduce to the corresponding (2, 2) mode data when the system is nonprecessing, but we treat nonprecessing cases differently by allowing the user to include only the (2,2) mode" 

Otherwise the reader will by like: If they reduce to the same thing, why do you have two things?

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    more details.
+
+    We currently only allow time-domain, nonprecessing waveforms. We
```
vijayvarma392, line 95: Get rid of ", nonprecessing"

vijayvarma392 (changes requested):

gw_eccentricity/gw_eccentricity.py
```diff
@@ -63,29 +63,52 @@ def measure_eccentricity(tref_in=None,
...
+    Throughout this documentation, we will refer to omega_gw and amp_gw. For
+    nonprecessing systems, these quantities reduce to the corresponding values
+    obtained using the (2, 2) mode, i. e., for nonprecessing systems,
+
+    amp_gw = amp22, phase_gw = phase22 and omega_gw = omega22
+
+    On the other hand, for precessing systems, we use Eq.(48) and (49) of
+    arXiv:1701.00550 to define amp_gw and phase_gw. amp_gw [phase_gw] is
+    defined using a symmetric [antisymmetric] combination of
+    amplitude [phase] of (2, 2) and (2, -2) mode.
```
vijayvarma392, line 85: Say coprecessing frame.

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

gw_eccentricity/gw_eccentricity.py
md-arif-shaikh: Done

vijayvarma392 (approved):

PR #220, opened 2024-09-17
support Precessing waveforms

md-arif-shaikh:
Changed a few lines to make it work with processing waveforms.


gw_eccentricity/eccDefinition.py
```diff
@@ -685,16 +682,23 @@ def get_amp_phase_omega_gw(self):
         system is nonprecessing.
         """
         if not self.precessing:
-            amp_gw, phase_gw, omega_gw = (self.dataDict["amplm"][(2, 2)],
-                                          self.dataDict["phaselm"][(2, 2)],
-                                          self.dataDict["omegalm"][(2, 2)])
+            amp_gw, phase_gw, omega_gw = (data_dict["amplm"][(2, 2)],
+                                          data_dict["phaselm"][(2, 2)],
+                                          data_dict["omegalm"][(2, 2)])
```
vijayvarma392, line 687: Was there are reason for changing from self.dataDict to data_dict?

md-arif-shaikh: Yes, we need to call this function inside the process_data_dict before we assign dataDict to self. There we need to provide the dataDict as an argument to this function which gives the amp_gw, phase_gw, omega_gw. Also see related #221 which I will address after this.

md-arif-shaikh: Can we merge this?

gw_eccentricity/eccDefinition.py
```diff
@@ -685,16 +682,23 @@ def get_amp_phase_omega_gw(self):
...
-            amp_gw = 0.5 * (self.dataDict[(2, 2)] + self.dataDict[(2, -2)])
-            phase_gw = 0.5 * (np.unwrap(np.angle(self.dataDict[(2, 2)]))
-                              - np.unwrap(np.angle(self.dataDict[(2, -2)])))
+            # check whether (2, -2) mode is provided.
+            for k in ["amplm", "phaselm"]:
+                if (2, -2) not in data_dict[k]:
+                    raise Exception(f"(2, -2) mode not found in {k}. For precessing"
+                             " systems, (2, -2) mode should be included in "
+                             "`dataDict`.")
+            amp_gw = 0.5 * (data_dict["amplm"][(2, 2)]
```
vijayvarma392, line 695: I've forgotten by now. Is the assumption that the user provides dataDict with waveform already in the coprecessing frame? If so, add a comment about that here, and that this needs to be handled more carefully in the future if we don't require that from the user.

md-arif-shaikh: Yes, that is correct. At this stage, we assume the modes are in coprecessing frame. When the next PR #222 to get modes in coprecessing frame is merged, we can add an option to rotate the modes internally by asking whether the modes are inertial or coprecessing frame. I will add these comments here.

md-arif-shaikh: comments added.

vijayvarma392 (approved):

PR #223, opened 2024-09-25
add rational fits

md-arif-shaikh:
Add `rational fit` as a fallback option to `Spline` when the interpolant of the `omega` extrema is not sufficiently smooth and its first derivative is nonmonotonic.

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -14,6 +14,8 @@ from .utils import time_deriv_4thOrder
 from .utils import interpolate
 from .utils import get_interpolant
 from .utils import get_default_spline_kwargs
+from .utils import get_rational_fit
+from .utils import get_default_rational_fit_kwargs
 from .utils import debug_message
 from .plot_settings import use_fancy_plotsettings, colorsDict, labelsDict
 from .plot_settings import figWidthsTwoColDict, figHeightsDict
@@ -310,9 +312,15 @@ class eccDefinition:
             get_default_spline_kwargs(),
             "spline_kwargs",
             "utils.get_default_spline_kwargs()")
+        self.rational_fit_kwargs = check_kwargs_and_set_defaults(
+            self.extra_kwargs["rational_fit_kwargs"],
+            get_default_rational_fit_kwargs(),
+            "rational_fit_kwargs",
+            "utils.get_default_rational_fit_kwargs()")
         self.available_averaging_methods \
             = self.get_available_omega_gw_averaging_methods()
         self.debug_level = self.extra_kwargs["debug_level"]
+        self.rational_fit_kwargs["verbose"] = True if self.debug_level >=1 else False
         self.debug_plots = self.extra_kwargs["debug_plots"]
         self.return_zero_if_small_ecc_failure = self.extra_kwargs["return_zero_if_small_ecc_failure"]
         # check if there are unrecognized keys in the dataDict
@@ -823,6 +831,7 @@ class eccDefinition:
         """Defaults for additional kwargs."""
         default_extra_kwargs = {
             "spline_kwargs": {},
+            "rational_fit_kwargs": {},
             "extrema_finding_kwargs": {},   # Gets overridden in methods like
                                             # eccDefinitionUsingAmplitude
             "debug_level": 0,
@@ -1174,6 +1183,53 @@ class eccDefinition:
             raise Exception(
                 f"Sufficient number of {extrema_type} are not found."
                 " Can not create an interpolant.")
+        
+    def rational_fit(self, x, y):
+        return get_rational_fit(x, y,
+                                rational_fit_kwargs=self.rational_fit_kwargs,
+                                check_kwargs=False)
+
+    def rational_fit_extrema(self, extrema_type="pericenters"):
+        if extrema_type == "pericenters":
+            extrema = self.pericenters_location
+        elif extrema_type == "apocenters":
+            extrema = self.apocenters_location
+        else:
+            raise Exception("extrema_type must be either "
+                            "'pericenrers' or 'apocenters'.")
+        if len(extrema) >= 2:
+            # assign degree
+            # TODO: Use an optimal degree based on the wavefrom length
+            self.rational_fit_kwargs["num_degree"],\
+                self.rational_fit_kwargs["denom_degree"] = 2, 2
+            rational_fit = self.rational_fit(self.t[extrema],
+                                             self.omega_gw[extrema])
+            t = self.t_for_checks.copy()
+            omega = rational_fit(t)
+            while self.check_domega_dt(t, omega, 1.0):
+                self.rational_fit_kwargs["num_degree"] -= 1
+                self.rational_fit_kwargs["denom_degree"] -= 1
+                debug_message(f"degree is lowered to {self.rational_fit_kwargs['num_degree']}",
+                              debug_level=self.debug_level,
+                              important=True)
+                rational_fit = self.rational_fit(self.t[extrema],
+                                                 self.omega_gw[extrema])
+                omega = rational_fit(t)
+            return rational_fit
+                
+        else:
+            raise Exception(
+                f"Sufficient number of {extrema_type} are not found."
+                " Can not create an interpolant.")
+
+    def check_domega_dt(self, t, omega, tol=1.0):
+        """Check first derivative of omega.
+
+        The first derivative of interpolant/fit of omega at the extrema
+        should be monotonically increasing.
+        """
+        domega_dt = np.gradient(omega, t[1] - t[0])
+        return any(domega_dt[1:]/domega_dt[:-1] < tol)
 
     def check_num_extrema(self, extrema, extrema_type="extrema"):
         """Check number of extrema.
@@ -1518,6 +1574,7 @@ class eccDefinition:
         self.t_apocenters = self.t[self.apocenters_location]
         self.tmax = min(self.t_pericenters[-1], self.t_apocenters[-1])
         self.tmin = max(self.t_pericenters[0], self.t_apocenters[0])
+        
         if self.domain == "frequency":
             # get the tref_in and fref_out from fref_in
             self.tref_in, self.fref_out \
@@ -1530,6 +1587,19 @@ class eccDefinition:
         self.t_for_checks = self.dataDict["t"][
             np.logical_and(self.dataDict["t"] >= self.tmin,
                            self.dataDict["t"] <= self.tmax)]
+
+        # Check monotonicity of the first derivative of omega_gw interpolant
+        # and build rational fit if necessary.
+        if (self.check_domega_dt(self.t_for_checks,
+                                 self.omega_gw_pericenters_interp(self.t_for_checks))
+            or
+            self.check_domega_dt(self.t_for_checks,
+                                 self.omega_gw_apocenters_interp(self.t_for_checks))):
+            debug_message("The spline interpolant through extrema has non "
+                          "monotonic time derivative. Trying rational fit instead.",
+                          debug_level=self.debug_level, important=True)
+            self.omega_gw_pericenters_interp = self.rational_fit_extrema("pericenters")
+            self.omega_gw_apocenters_interp = self.rational_fit_extrema("apocenters")
 
         # Sanity checks
         # check that fref_out and tref_out are of the same length
--- a/gw_eccentricity/utils.py
+++ b/gw_eccentricity/utils.py
@@ -1,6 +1,7 @@
 """Useful functions for the project."""
 import numpy as np
 import argparse
+from polyrat import StabilizedSKRationalApproximation
 from scipy.interpolate import InterpolatedUnivariateSpline
 from scipy.interpolate import PchipInterpolator
 import warnings
@@ -315,6 +316,37 @@ def get_interpolant(oldX,
         raise ValueError(f"Unknown interpolator {interpolator}. Must be one"
                          " of ['spline', 'monotonic_spline']")
     return interpolant
+
+
+def get_default_rational_fit_kwargs():
+    """Get default kwargs for rational fit."""
+    default_rational_fit_kwargs = {
+        "num_degree": None,
+        "denom_degree": None,
+        "norm": 2,
+        "maxiter": 20,
+        "verbose": False,
+        "xtol": 1e-07,
+    }
+    return default_rational_fit_kwargs
+
+
+def get_rational_fit(x, y, rational_fit_kwargs=None, check_kwargs=True):
+    """Get rational fit for the data set (x, y).
+
+    We use `polyrat.StabilizedSKRationalApproximation` to obtain
+    rational fit to the data.
+    """
+    if check_kwargs:
+        rational_fit_kwargs = check_kwargs_and_set_defaults(
+            rational_fit_kwargs,
+            get_default_rational_fit_kwargs(),
+            "rational_fit_kwargs",
+            "utils.get_default_rational_fit_kwargs"
+        )
+    rat = StabilizedSKRationalApproximation(**rational_fit_kwargs)
+    rat.fit(x.reshape(-1, 1), y)
+    return lambda x: rat(x.astype('float64').reshape(-1, 1))
 
 
 def debug_message(message, debug_level, important=True,
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,8 @@ setuptools.setup(
         'scipy',
         'h5py',
         'lalsuite',
-        'sxs'
+        'sxs',
+        'polyrat'
     ],
     classifiers=[
         "Intended Audience :: Science/Research",


md-arif-shaikh:
This PR is quite long, so I want to highlight the main changes

- utils.py: general rational fit functions added here
- eccDefinition.py: rational fit functions for fit to the omega_gw_extrema added. This is the most important function to be reviewed.

Other changes including the tests are straightforward. I Had to change a few things to make the tests pass because we have changed the default to `rational_fit`. More changes are to be made to the tests but that can be done later in a separate PR.

md-arif-shaikh:
Hi @vijayvarma392, this PR is now ready to be reviewed. I have summarized the main changes [here](https://github.com/vijayvarma392/gw_eccentricity/pull/223#issuecomment-2439871216). I think at this point, I would like to have your comments for making further changes.

gw_eccentricity/eccDefinition.py
```diff
@@ -174,6 +176,11 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                 omega_gw_apocenters(t).
                 Defaults are set using utils.get_default_spline_kwargs
 
+            rational_fit_kwargs: dict
+                Dictionary of arguments to be passed to the rational
+                fit function. Defaults are set using
+                `utils.get_default_rational_fit_kwargs`
+
```
vijayvarma392, line 183: You should indicate here already that only one of spline_kwargs or rational_fit_kwargs are allowed.
Question: Would it be better to have interp_kwargs, which has a key 'interp_type' that can be 'spline' or 'rational' fit? In the future, if we add extra options like Toni's PN inspired fits, that would make it easier.
This is more of a question for you rather than a suggestion because I don't know if that will make the code more complicated than it needs to be. Either way, needs to be clear that only one is allowed.

gw_eccentricity/eccDefinition.py
```diff
@@ -272,6 +279,29 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                Specifies the method used to build the interpolations for 
+                `omega_gw_pericenters_interp(t)` or `omega_gw_apocenters_interp(t)`.
+                The available options are:
+
+                - `spline`: Uses `scipy.interpolate.InterpolatedUnivariateSpline`.
+                    - Best suited for cleaner data, such as when waveform modes are generated
+                        using models like SEOB or TEOB.
+                    - Faster to construct and evaluate.
+                    - Since it fits through every data point, it may exhibit oscillatory 
+                        behavior, particularly near the merger.
```
vijayvarma392, line 293: near the merger, especially for noisy NR data.

gw_eccentricity/eccDefinition.py
```diff
@@ -272,6 +279,29 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+        
+                - `rational_fit`: Uses `polyrat.StabilizedSKRationalApproximation`.
+                    - Can handle both clean and noisy data, e.g., waveform
+                        modes from numerical simulations.
+                    - Better monotonic behaviour, particularly near the merger.
+                    - Significantly slower compared to the `spline` method. This is because
+                        finding optimal numerator and denominator degree needs several iterations
```
vijayvarma392, line 300: Can you say a rough number instead of "Significantly slower"?

gw_eccentricity/eccDefinition.py
```diff
@@ -310,9 +340,21 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+            "utils.get_default_rational_fit_kwargs()")
         self.available_averaging_methods \
             = self.get_available_omega_gw_averaging_methods()
         self.debug_level = self.extra_kwargs["debug_level"]
+        # set verbose to debug_level. If verbose is True, then it prints information
+        # of each iteration for rational fits for a given degree.
+        self.rational_fit_kwargs["verbose"] = self.debug_level
+        # keep history of rational fit degree and nonmonotonicity of the corresponding fits
+        self.rational_fit_nonmonotonicity_history = {"pericenters": {}, "apocenters": {}}
+        # Update the degrees used to construct the final fits
```
vijayvarma392, line 356: Should the comment be: This is used to keep track and update the degrees used to construct the final fits?

gw_eccentricity/eccDefinition.py
```diff
@@ -310,9 +340,21 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+            self.extra_kwargs["rational_fit_kwargs"],
+            get_default_rational_fit_kwargs(),
+            "rational_fit_kwargs",
+            "utils.get_default_rational_fit_kwargs()")
         self.available_averaging_methods \
             = self.get_available_omega_gw_averaging_methods()
         self.debug_level = self.extra_kwargs["debug_level"]
+        # set verbose to debug_level. If verbose is True, then it prints information
+        # of each iteration for rational fits for a given degree.
+        self.rational_fit_kwargs["verbose"] = self.debug_level
```
vijayvarma392, line 353: Why not just call it debug_level?

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
             raise Exception(
                 f"Sufficient number of {extrema_type} are not found."
                 " Can not create an interpolant.")
+        
+    def get_rat_fit(self, x, y):
```
vijayvarma392, line 1232: hmm, don't like the name...come up with a way to not call the two things get_rat_fit and get_rational_fit? Easily confused.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+                                rational_fit_kwargs=self.rational_fit_kwargs,
+                                check_kwargs=False)
+
+    def rational_fit(self, x, y, description=None):
+        """Get rational fit with adaptive numerator and denominator degree.
+
+        This function ensures that the rational fit uses the optimal degree for
+        the numerator and denominator. It first checks for nonmonotonicity in the
+        fit's derivative and lowers the degrees if necessary. Only if the initial
+        degree was monotonic (no reduction needed), it attempts to increase the
```
vijayvarma392, line 1251: "Only if the initial degree was monotonic" --> "Only if the fit for the initial degree was monotonic"

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+        degrees to find a higher-degree fit. If increasing the degrees causes
+        nonmonotonicity, it reverts to the previous valid monotonic fit.
+        """
+        # make sure that description is not None. A description is needed to
+        # update the optimal values of the numerator and denominator degrees
+        # used to build the final rational fit
+        if description is None:
```
vijayvarma392, line 1258: description --> data_name?

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+
+    def rational_fit(self, x, y, description=None):
+        """Get rational fit with adaptive numerator and denominator degree.
+
+        This function ensures that the rational fit uses the optimal degree for
+        the numerator and denominator. It first checks for nonmonotonicity in the
+        fit's derivative and lowers the degrees if necessary. Only if the initial
+        degree was monotonic (no reduction needed), it attempts to increase the
+        degrees to find a higher-degree fit. If increasing the degrees causes
+        nonmonotonicity, it reverts to the previous valid monotonic fit.
```
vijayvarma392, line 1253: Say something about how the initial degrees should be set? Through self.rational_fit_kwargs?

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+        # The optimal degrees for the numerator and denominator changes depending on the
+        # eccentricity and duration of the waveform. We set these values by counting the
+        # number of approximate orbits and using some prior mapping between number of orbits
+        # and the optimal degrees. The mapping is done via `get_approximate_degree_for_rational_fit`.
+        # Usually this is a good starting point but not the optimal one
+        # as it varies as a function of eccentricity which we do not know apriori. The optimal
+        # value is found by doing some iterations.
+        if not (self.rational_fit_kwargs["num_degree"] and self.rational_fit_kwargs["denom_degree"]):
+            self.rational_fit_kwargs["num_degree"], self.rational_fit_kwargs["denom_degree"] \
+                = self.get_approximate_degree_for_rational_fit()
```
vijayvarma392, line 1272: Does this still work if the user gives only one of num_degree and denom_degree?

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+            num_degree, denom_degree = 3, 3
+        elif (approximate_num_orbits > 50) and (approximate_num_orbits <= 100):
+            num_degree, denom_degree = 4, 4
+        elif (approximate_num_orbits > 100) and (approximate_num_orbits <= 150):
+            num_degree, denom_degree = 5, 5
+        elif (approximate_num_orbits > 150) and (approximate_num_orbits <= 200):
+            num_degree, denom_degree = 6, 6
+        else:
+            num_degree, denom_degree = 5 + int(np.log10(approximate_num_orbits)), 5 + int(np.log10(approximate_num_orbits))
+        return num_degree, denom_degree
```
vijayvarma392, line 1361: If you check for approximate_num_orbits <= 5 once, you don't need to check for approximate_num_orbits > 5 again in elif. Something like:
```
if orbits <= 5:
    bla
elif orbits <= 20:
```

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+        # and the optimal degrees. The mapping is done via `get_approximate_degree_for_rational_fit`.
+        # Usually this is a good starting point but not the optimal one
+        # as it varies as a function of eccentricity which we do not know apriori. The optimal
+        # value is found by doing some iterations.
+        if not (self.rational_fit_kwargs["num_degree"] and self.rational_fit_kwargs["denom_degree"]):
+            self.rational_fit_kwargs["num_degree"], self.rational_fit_kwargs["denom_degree"] \
+                = self.get_approximate_degree_for_rational_fit()
+
+        rat_fit = self.get_rat_fit(x, y)
+        t = np.arange(x[0], x[-1], self.t[1] - self.t[0])
```
vijayvarma392, line 1275: best not to call this t. There is already a self.t, and the natural thing to do is to assume they are the same.

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+        # save the degrees for checks at each step of iterations for finding the optimal degrees
+        old_num_degree = self.rational_fit_kwargs["num_degree"]
+        old_denom_degree = self.rational_fit_kwargs["denom_degree"]
+
+        # use a flag to know whether the initial degrees produced monotonic
+        # time derivative of the fits or it was required to be reduced.
+        degrees_were_reduced = False
+
+        # Check for nonmonotonicity and lower degrees if needed
+        while self.check_if_domega_dt_is_nonmonotonic(t, rat_fit(t), 1.0, description):
```
vijayvarma392, line 1285: Should the function be called something other than `check_if_domega_dt_is_nonmonotonic`? Isn't the data allowed to be things other than omega?

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+                              "and cannot be lowered further.",
+                              debug_level=self.debug_level, important=False)
+                break
+
+            debug_message(f"Lowering degrees to num_degree={self.rational_fit_kwargs['num_degree']}, "
+                          f"denom_degree={self.rational_fit_kwargs['denom_degree']} and retrying.",
+                          debug_level=self.debug_level, important=False)
+
+            rat_fit = self.get_rat_fit(x, y)
+        # If no degrees were reduced, try increasing the degree for better fit
```
vijayvarma392, line 1302: for a better fit

gw_eccentricity/eccDefinition.py
```diff
@@ -1184,6 +1228,180 @@ def interp_extrema(self, extrema_type="pericenters"):
...
+            last_monotonic_denom_degree = old_denom_degree
+
+            while not self.check_if_domega_dt_is_nonmonotonic(t, rat_fit(t), 1.0, description):
+                # Increase the degrees for both numerator and denominator
+                new_num_degree = self.rational_fit_kwargs["num_degree"] + 1
+                new_denom_degree = self.rational_fit_kwargs["denom_degree"] + 1
+                self.rational_fit_kwargs["num_degree"], self.rational_fit_kwargs["denom_degree"] \
+                    = new_num_degree, new_denom_degree
+                new_rat_fit = self.get_rat_fit(x, y)
+                if self.check_if_domega_dt_is_nonmonotonic(t, new_rat_fit(t), 1.0, description):
```
vijayvarma392, line 1315: Looks like this check happens twice for the same fit in some cases? Maybe rework the code so that that doesn't happen?

gw_eccentricity/eccDefinition.py
vijayvarma392: How did you arrive at these choices? Add some comments about how you got there and how important these particular choices are.

gw_eccentricity/eccDefinition.py
```diff
@@ -1597,6 +1814,33 @@ def measure_ecc(self, tref_in=None, fref_in=None):
         # return measured eccentricity, mean anomaly and reference time or
         # frequency where these are measured.
         return self.make_return_dict_for_eccentricity_and_mean_anomaly()
+    
+    def build_omega_gw_extrema_interpolants(self):
```
vijayvarma392, line 1818: Needs a docstring explaining what it's doing and why.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: We use spline kwargs for things other than interpolating the omega_gw extrema. For example, we interpolate zeroecc data using spline. So providing just rational_fit kwargs won't work because rational_fit is only good for interpolating monotonic trends.

md-arif-shaikh: Basically we will need to make all interpolation thing work with a single method if we want the user to provide only one of them. Here, we wanted to use rational fit for the omega envelope mainly.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Ok, I have changed it to `check_if_first_derivative_is_not_strictly_monotonic`

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: corrected. These are based on limited investigations. It needs a lot more tests to set more optimal values which I intend to do once this is merged. I have noted this using TODO at the top

gw_eccentricity/eccDefinition.py
md-arif-shaikh: No, I have changed this part of the code so that it does the following: If both are None, i.e., unspecified only then use the function to set the approximate degree to both num and denom. If any one of the degrees is given, use it to set the other.

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Okay, so I am using `rational_fit_nonmonotonicity_history` to keep track of what degrees are being iterated over and the corresponding behaviour, i.e., is it nonmonotonic? And using `rational_fit_degrees` to store the final degree used to build the fits. Instead of using "update the degrees", I changed it to "store degrees" to better convey the purpose.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: "verbose" is the keyword used in the routine for rational fit, so I have kept that the same and instead assign value to it using `debug_level`

gw_eccentricity/eccDefinition.py
md-arif-shaikh: changed `get_rat_fit` to `get_rational_fit_wrapper`

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: What do you suggest?

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Rephrased it.

gw_eccentricity/eccDefinition.py
```diff
@@ -205,13 +208,30 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         extra_kwargs: dict
             A dictionary of any extra kwargs to be passed. Allowed kwargs
             are:
-            spline_kwargs: dict
+            special_interp_kwargs_for_extrema: dict
+                A dictionary with a single key matching the current
+                `omega_gw_extrema_interpolation_method`. See under
```
vijayvarma392, line 213: Why the word "current" here?

gw_eccentricity/eccDefinition.py
```diff
@@ -205,13 +208,30 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                  `utils.get_default_rational_fit_kwargs`
+
+            general_interp_kwargs: dict
                 Dictionary of arguments to be passed to the spline
                 interpolation routine
                 (scipy.interpolate.InterpolatedUnivariateSpline) used to
-                compute quantities like omega_gw_pericenters(t) and
-                omega_gw_apocenters(t).
-                Defaults are set using utils.get_default_spline_kwargs
+                interpolate data other than the omega_gw extrema.
```
vijayvarma392, line 233: Based on this description, how about renaming special_interp_kwargs_for_extrema --> special_interp_kwargs_for_omega_gw_extrema?

gw_eccentricity/eccDefinition.py
```diff
@@ -303,6 +323,31 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                 eccentricity is the cause, and set the returned eccentricity
                 and mean anomaly to zero.
                 USE THIS WITH CAUTION!
+
+            omega_gw_extrema_interpolation_method : str, default="rational_fit"
```
vijayvarma392, line 327: Move this above, to before special_interp_kwargs_for_extrema?

gw_eccentricity/eccDefinition.py
```diff
@@ -303,6 +323,31 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+        
+                - `rational_fit`: Uses `polyrat.StabilizedSKRationalApproximation`.
+                    - Can handle both clean and noisy data, e.g., waveform
+                        modes from numerical simulations.
+                    - Better monotonic behaviour, particularly near the merger.
+                    - Can be slower compared to the `spline` method.
+                      This is because finding optimal numerator and denominator
+                      degree may need several iterations.
+                    - Can suppress pathologies in the waveform that might be
+                      visible with `spline`.
```
vijayvarma392, line 348: visible --> seen

gw_eccentricity/eccDefinition.py
```diff
@@ -303,6 +323,31 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                      oscillatory behavior, particularly near the merger,
+                      especially for noisy NR data.
+        
+                - `rational_fit`: Uses `polyrat.StabilizedSKRationalApproximation`.
+                    - Can handle both clean and noisy data, e.g., waveform
+                        modes from numerical simulations.
+                    - Better monotonic behaviour, particularly near the merger.
+                    - Can be slower compared to the `spline` method.
+                      This is because finding optimal numerator and denominator
+                      degree may need several iterations.
```
vijayvarma392, line 346: Is this optimization over degree documented somewhere else? If so, point to that here.

gw_eccentricity/eccDefinition.py
```diff
@@ -981,11 +1054,94 @@ def get_default_extrema_finding_kwargs(self, width):
...
+    
+    def get_default_special_interp_kwargs_for_extrema(self):
+        """Get default kwargs to be passed to extrema interpolating method.
+        
+        Accurate interpolation of omega_gw extrema is crucial for obtaining
+        monotonic eccentricity evoultion with time. Depending on the
+        interpolation method, this function returns the default kwargs to be
+        passed to the interpolating function. 
+        """
+        allowed_methods = self.extra_kwargs["special_interp_kwargs_for_extrema"].keys()
```
vijayvarma392, line 1066: Confused. Why are allowed_methods in self.extra_kwargs["special_interp_kwargs_for_extrema"]?

gw_eccentricity/eccDefinition.py
```diff
@@ -981,11 +1054,94 @@ def get_default_extrema_finding_kwargs(self, width):
...
+            raise Exception("Unknown omega_gw extrema interpolation method "
+                            f"`{self.omega_gw_extrema_interpolation_method}`. "
+                            f"Allowed methods are {allowed_methods}")
+        return kwargs
+    
+    def check_special_interp_kwargs_for_extrema(self, extra_kwargs):
+        """Check extrema interp kwargs provided in extra_kwargs.
+        
+        The `special_interp_kwargs_for_extrema` should have exactly one
+        key matching the current `omega_gw_extrema_interpolation_method`.
```
vijayvarma392, line 1081: If that is the case, why even leave special_interp_kwargs_for_extrema as an option for the user?

gw_eccentricity/eccDefinition.py
```diff
@@ -1285,33 +1442,217 @@ def get_good_extrema(self, pericenters, apocenters,
...
+    def get_rational_fit_for_extrema(self, x, y, data_name=None):
+        """Get rational fit with adaptive numerator and denominator degree.
+
+        This function ensures that the rational fit uses the optimal degree for
+        the numerator and denominator. It first checks for nonmonotonicity in
+        the fit's derivative and lowers the degrees if necessary. Only if the
+        fit for the initial degree was monotonic (no reduction of degree
+        needed), it attempts to increase the degrees to find a higher-degree
+        fit. If increasing the degrees causes nonmonotonicity, it reverts to
+        the previous valid monotonic fit.
```
vijayvarma392, line 1454: A bit more explanation would be nice. Maybe describe the form of the rational polinomial and say whether you increase/decrease the degree of denominator and numerator at the same time. Also mention what the minimum and  maximum degrees are.

gw_eccentricity/eccDefinition.py
```diff
@@ -1285,33 +1442,217 @@ def get_good_extrema(self, pericenters, apocenters,
...
+        # predefined mapping from orbit count to optimal degrees, provided by
+        # `get_approximate_degree_for_rational_fit`. 
+        
+        # While this provides a good initial estimate, it may not be optimal,
+        # as the ideal degree can change based on the eccentricity which we do
+        # not know apriori. The true optimal values are determined through
+        # further iterations.
+        if self.special_interp_kwargs_for_extrema["num_degree"] is None \
+            and self.special_interp_kwargs_for_extrema["denom_degree"] is None:
+            self.special_interp_kwargs_for_extrema["num_degree"], \
```
vijayvarma392, line 1490: Mention in a comment what this code is doing: If only one of num_degree/denom_degree is given, it sets the other one to match it.

gw_eccentricity/eccDefinition.py
```diff
@@ -1285,33 +1442,217 @@ def get_good_extrema(self, pericenters, apocenters,
...
+            rat_fit = get_rational_fit(
+                x, y,
+                rational_fit_kwargs=self.special_interp_kwargs_for_extrema,
+                check_kwargs=False)
+            fit_is_nonmonotonic \
+                = self.check_if_first_derivative_is_not_strictly_monotonic(
+                x_test, rat_fit(x_test), 1.0, data_name)
+        # If fit with initial degree is monotonic, try increasing the degree
+        # for a better fit
+        if not fit_is_nonmonotonic:
```
vijayvarma392, line 1543: Would the code be easier to follow if it's like:
if fit_is_nonmonotonic:
   do something
else:
   do something else

gw_eccentricity/eccDefinition.py
```diff
@@ -1285,33 +1442,217 @@ def get_good_extrema(self, pericenters, apocenters,
...
+        # since the relation between the final degrees and number of
+        # cycles is not very simple, and are meant only for a
+        # reasonable guess for the initial degrees which is then
+        # adapted to get the final degrees.
+        
+        # assign degree based on the number of extrema if user provided
+        # degree is None.
+        approximate_num_orbits = max(len(self.pericenters_location),
+                                     len(self.apocenters_location))
+        if approximate_num_orbits <= 20:
```
vijayvarma392, line 1608: Maybe make a wiki page with the investigations you did and link that here as motivation for why you use these numbers?

gw_eccentricity/eccDefinition.py
```diff
@@ -1749,6 +2102,29 @@ def measure_ecc(self, tref_in=None, fref_in=None):
...
         return self.make_return_dict_for_eccentricity_and_mean_anomaly()
+    
+    def check_omega_gw_extrema_interpolants(self):
+        """
+        check monotonicity of the omega_gw extrema interpolants.
+        """
+        # Verify the monotonicity of the first derivative of the omega_gw
+        # interpolant with spline.
+        if self.extra_kwargs["omega_gw_extrema_interpolation_method"] \
+            == "spline":
```
vijayvarma392, line 2113: Why only for spline? If it's because you already do this test for rational_fits, should this check happen within the method that builds the spline? Otherwise, the code seems confusing. 

Similarly, I'm not a fan of if conditions like this
https://github.com/vijayvarma392/gw_eccentricity/blob/b8872a1e4bb44f74287583007d995fb3003eaeae/gw_eccentricity/eccDefinition.py#L1626
Is there a way to just call this test from within the function that does that rational fit?

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Removed

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Fixed

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Linked a wiki page.

gw_eccentricity/load_data.py
```diff
@@ -1019,7 +1019,7 @@ def check_sxs_data_dir(origin, **kwargs):
         f"file named `Strain_N{kwargs['extrap_order']}` since `extrap_order` "
         f"is {kwargs['extrap_order']}."}
     if any([kwargs["include_zero_ecc"], kwargs["include_params_dict"],
-            not kwargs["keep_memory"]]):
+            not kwargs.get("keep_memory", True)]):
```
vijayvarma392, line 1022: Hmm, are we keeping memory by default? Maybe better to do the opposite?

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
```diff
@@ -205,13 +208,49 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                    - Can suppress pathologies in the waveform that might be
+                      seen with `spline`.
+
+                Default value: `"rational_fit"`.
+
+            special_interp_kwargs_for_omega_gw_extrema: dict
+                A dictionary of kwargs for `omega_gw_extrema_interpolation_method`.
+                Currently, the available `omega_gw_extrema_interpolation_method` and
+                the corresponding defaults for `special_interp_kwargs_for_omega_gw_extrema`
+                are:
```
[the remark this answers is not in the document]
md-arif-shaikh: Done

gw_eccentricity/load_data.py
md-arif-shaikh: No, the default value is `False`. So by default `not kwargs.get("keep_memory", True)])` will be True and therefore one requires to give path to metadata. The default value (the secoond arg `True`) to `kwargs.get()` is to ensure that if someone is loading old format sxs waveform where "keep_memory" is not a kwarg, `not kwargs.get()` will evaluate to False, meaning no requirement to provide metadata file.

PR #225, opened 2024-10-28
Add option to remove memory from sxs waveforms

md-arif-shaikh:
Removing memory may give better results for high mass-ration, and eccentric precessing systems. This PR adds an option for `sxs` waveforms.

### DIFF
--- a/gw_eccentricity/load_data.py
+++ b/gw_eccentricity/load_data.py
@@ -8,6 +8,7 @@ import lal
 import lalsimulation as lalsim
 import warnings
 import json
+import spherical_functions as sf
 from copy import deepcopy
 from .utils import peak_time_via_quadratic_fit
 from .utils import amplitude_using_all_modes
@@ -90,8 +91,9 @@ def get_load_waveform_defaults(origin="LAL"):
                        "zero_ecc_approximant",
                        "num_orbits_to_remove_as_junk",
                        "mode_array",
-                       "extrap_order",
-                       "remove_memory"]
+                       "extrap_order"]
+        if origin == "SXSCatalog":
+            kwargs_list.append("remove_memory")
         return make_a_sub_dict(get_defaults_for_nr(), kwargs_list)
     # for waveforms in LVCNR format file using recommended function in LALSuite
     elif origin == "LVCNR":
@@ -532,8 +534,7 @@ def get_defaults_for_nr():
         If True, remove memory contribution from the waveform modes.
         This will require metadata file to find t_relax which is used
         to start the integration for computing memory contribution.
-        NOTE: This is currently implemented ony for sxs catalog formatted
-        waveforms.
+        NOTE: This is used only in the newer sxs catalog formatted waveforms
     """
     return {"filepath": None,
             "data_dir": None,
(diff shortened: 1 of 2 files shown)


gw_eccentricity/load_data.py
```diff
@@ -526,6 +534,14 @@ def get_defaults_for_nr():
     extrap_order: int
         Extrapolation order to use for loading the waveform data.
         NOTE: This is used only for sxs catalog formatted waveforms.
+
+    remove_memory: bool
```
vijayvarma392, line 538: I wonder if it's best to remove memory by default. So, keep_memory?

gw_eccentricity/load_data.py
```diff
@@ -1093,9 +1127,9 @@ def get_modes_dict_from_sxs_catalog_old_format(**kwargs):
             time = mode_data[:, 0]
             t = np.arange(time[0], time[-1], kwargs["deltaTOverM"])
         hlm = mode_data[:, 1] + 1j * mode_data[:, 2]
-        amp_interp = interpolate(t, time, np.abs(hlm))
-        phase_interp = interpolate(t, time, -np.unwrap(np.angle(hlm)))
-        hlm_interp = amp_interp * np.exp(-1j * phase_interp)
+        real_interp = interpolate(t, time, np.real(hlm))
+        imag_interp = interpolate(t, time, np.imag(hlm))
+        hlm_interp = real_interp + 1j * imag_interp
```
vijayvarma392, line 1132: I forget, what is the motivation for this change?

gw_eccentricity/load_data.py
```diff
@@ -1109,6 +1143,28 @@ def get_modes_dict_from_sxs_catalog_format(**kwargs):
     # get the waveform object
     waveform = sxs.rpdmb.load(
```
vijayvarma392, line 1144: Nowadays, you can just do sxs.load

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
md-arif-shaikh: Do you suggest `keep_memory` with default set to `True`?

gw_eccentricity/load_data.py
md-arif-shaikh: In the extreme precession case, intepolating amp and phase on smaller timestep was introducing some artefacts where as real and imaginary was not having those issues.

gw_eccentricity/load_data.py
md-arif-shaikh: Okay, I will use that.

gw_eccentricity/load_data.py
vijayvarma392: No, with default False, so that by default no memory.

gw_eccentricity/load_data.py
md-arif-shaikh: moved discussion to slack

md-arif-shaikh: Added comments

gw_eccentricity/load_data.py
```diff
@@ -824,6 +843,11 @@ def load_sxs_catalogformat(**kwargs):
         to locate the strain file. This function will seek a file named
         `Strain_N{extrap_order}.h5` in the `data_dir`.
 
+    keep_memory: bool
+        If False, remove memory contribution from the waveform modes.
+        This will require metadata file to find t_relax which is used
+        to start the integration for computing memory contribution.
```
vijayvarma392, line 849: Should it say what the default is?

gw_eccentricity/load_data.py
```diff
@@ -526,6 +535,14 @@ def get_defaults_for_nr():
     extrap_order: int
         Extrapolation order to use for loading the waveform data.
         NOTE: This is used only for sxs catalog formatted waveforms.
+
+    keep_memory: bool
+        If False, remove memory contribution from the waveform modes.  This will
+        require metadata file to find t_relax which is used to start the
+        integration for computing memory contribution.  NOTE: This can be used
+        only in the newer sxs catalog formatted waveforms with
+        origin=`SXSCatalog`
```
vijayvarma392, line 544: Does the code raise a helpful Exception if this is used with catalogs other than SXSCatalog?

vijayvarma392 (changes requested):

gw_eccentricity/load_data.py
md-arif-shaikh: It should by design since it is included only in the list of kwargs for SXSCatalog. I will share an example case.

gw_eccentricity/load_data.py
md-arif-shaikh: Earlier we decided that we are not going to mention the default for each kwarg here and instead refer to `load_data.get_load_waveform_defaults('SXSCatalog')` for getting the defaults. For example at the of of the do string under parameters, we refer to this function. The idea was to set the defaults for NR at a single place because some of these are shared by different waveform origins.

vijayvarma392: Ok, I'll leave it to you, just make sure it would be easy for a reader to understand, and feel free to veto old decisions if they no longer make sense.

gw_eccentricity/load_data.py
md-arif-shaikh: Here is the error message if `keep_memory` is used, for example, with `SXSCatalog_old`
image

gw_eccentricity/load_data.py
md-arif-shaikh: I don't have a strong opinion about changing things as of now. But we can consider changing the current setup later, possibly in a separate PR.

md-arif-shaikh:
Merge?

PR #226, opened 2024-11-10
add option for frame to rotate modes internally if required

md-arif-shaikh:
Adds an argument `frame` which takes either of the two values `inertial` or `coprecessing`. If the system is `precessing` and `frame='inertial'`, the modes are rotated internally to get the dataDict in the coprecessing frame. This completes the extension of `gw_eccentricity` to precessing system.

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -8,6 +8,7 @@ https://github.com/vijayvarma392/gw_eccentricity/wiki/Adding-new-eccentricity-de
 
 import numpy as np
 import matplotlib.pyplot as plt
+import warnings
 from .load_data import get_coprecessing_data_dict
 from .utils import peak_time_via_quadratic_fit, check_kwargs_and_set_defaults
 from .utils import amplitude_using_all_modes
@@ -169,10 +170,11 @@ class eccDefinition:
                   directly via `dataDict`.
                 - Set `frame="inertial"` and provide the inertial frame modes
                   via `dataDict`. In this case, the modes in `dataDict` are
-                  rotated internally (see `rotate_modes`) before further
-                  computation. To get the coprecessing modes from the inertial
-                  modes accurately, `dataDict` must include all modes for
-                  `ell=2`, i.e., (2, -2), (2, -1), (2, 0), (2, 1) and (2, 2).
+                  rotated internally (see `transform_inertial_to_coprecessing`)
+                  before further computation. To get the coprecessing modes
+                  from the inertial modes accurately, `dataDict` must include
+                  all modes for `ell=2`, i.e., (2, -2), (2, -1), (2, 0), (2, 1)
+                  and (2, 2).
 
             For nonprecessing systems, the inertial and coprecessing frames are
             equivalent, so there is no distinction. For nonprecessing systems,
@@ -621,7 +623,18 @@ class eccDefinition:
         # frame, rotate the modes and obtain the corresponding modes in the
         # coprecessing frame
         if self.precessing is True and self.frame == "inertial":
-            dataDict = self.rotate_modes(dataDict)
+            warnings.warn(
+                f"The system is precessing but the modes are provided in "
+                f"the {self.frame} frame. Transforming the modes from"
+                f" the {self.frame} frame to the coprecessing frame and "
+                "updating `self.frame` to `coprecessing`.")
+            dataDict = self.transform_inertial_to_coprecessing(dataDict)
+            # transform the zeroecc modes as well if provided in dataDict
+            if "hlm_zeroecc" in dataDict or "amplm_zeroecc" in dataDict:
+                dataDict = self.transform_inertial_to_coprecessing(
+                    dataDict, suffix="_zeroecc")
+            # Now that the modes are in the coprecessing frame, update frame
+            self.frame = "coprecessing"
         # Create a new dictionary that will contain the data necessary for
         # eccentricity measurement.
         newDataDict = {}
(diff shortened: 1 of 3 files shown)


md-arif-shaikh:
@vijayvarma392  this PR is ready to review now. As described at the [top](https://github.com/vijayvarma392/gw_eccentricity/pull/226#issue-2647075576), it adds an option to rotate modes internally.

gw_eccentricity/eccDefinition.py
```diff
@@ -156,12 +158,42 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                  via `dataDict`. In this case, the modes in `dataDict` are
+                  rotated internally (see `rotate_modes`) before further
+                  computation. To get the coprecessing modes from the inertial
+                  modes accurately, `dataDict` must include all modes for
+                  `ell=2`, i.e., (2, -2), (2, -1), (2, 0), (2, 1) and (2, 2).
+
+            For nonprecessing systems, the inertial and coprecessing frames are
+            equivalent, so there is no distinction.
+
+            Default is `False`, indicating the system is nonprecessing.
```
vijayvarma392, line 180: For clarity, say that for nonprecessing, it's sufficient to include only the (2,2) mode.

gw_eccentricity/eccDefinition.py
```diff
@@ -156,12 +158,42 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
...
+                - "coprecessing": The modes in `dataDict` are in the
+                  coprecessing frame.
+
+            If the system is precessing (`precessing=True`) and
+            `frame="inertial"`, the modes in `dataDict` are rotated into the
+            coprecessing frame for further computation. If
+            `frame="coprecessing"` with a precessing system, the modes in
+            `dataDict` are expected to be already in the coprecessing frame.
+
+            Default value is "inertial".
```
vijayvarma392, line 196: Some of the description seems repetitive with the above, maybe take another pass to see if that can be avoided/reduced/simplified?

gw_eccentricity/eccDefinition.py
```diff
@@ -274,6 +306,13 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
                 USE THIS WITH CAUTION!
         """
         self.precessing = precessing
+        self.frame = frame
+        # check if frame makes sense.
+        self.available_frames = ["inertial", "coprecessing"]
```
vijayvarma392, line 311: Does `available_frames` needs to be in self? If not, don't put it in self, just makes self bloated, and makes it harder to find things you really need in self if you try to tab complete in a notebook.

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Done

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Rephrased to avoid repetition.

gw_eccentricity/eccDefinition.py
md-arif-shaikh: removed it from `self`

gw_eccentricity/eccDefinition.py
```diff
@@ -579,6 +617,11 @@ def process_data_dict(self,
                 "1. 'hlm' OR \n"
                 "2. 'amplm' and 'phaselm'\n"
                 "But not both 1. and 2. at the same time."))
+        # if the system is precessing and the modes are given in the inertial
+        # frame, rotate the modes and obtain the corresponding modes in the
+        # coprecessing frame
+        if self.precessing is True and self.frame == "inertial":
+            dataDict = self.rotate_modes(dataDict)
```
vijayvarma392, line 624: Once you do this, self.frame will be inertial, but dataDict will be in coprecessing?
Seeing as you do want to keep frame in self, why not update it at this point?
Also, `rotate_modes` seems too generic, how about `transform_inertial_to_coprecessing`?

gw_eccentricity/eccDefinition.py
```diff
@@ -654,11 +697,97 @@ def process_data_dict(self,
...
+            # Make a dictionary with a key "hlm" and pass the zeroecc modes 
+            # dict via it. Similary for the times using "t".
+            intertial_zeroecc_data_dict = {"t": data_dict["t_zeroecc"],
+                                           "hlm": data_dict["hlm_zeroecc"]}
+            coprecessing_zeroecc_data_dict = get_coprecessing_data_dict(
+                intertial_zeroecc_data_dict)
+            # update the data_dict by replacing the inertial frame zeroecc
+            # modes with coprecessing frame zeroecc modes.
+            data_dict.update(
+                {"hlm_zeroecc": coprecessing_zeroecc_data_dict["hlm"]})
```
vijayvarma392, line 740: Instead of all of this, what if you add a `tag` option to get_coprecessing_data_dict, that defaults to an empty string `""`, but you can pass `"_zeroecc"` if needeed?

gw_eccentricity/eccDefinition.py
```diff
@@ -654,11 +697,97 @@ def process_data_dict(self,
...
+                    inertial_zeroecc_data_dict)
+                data_dict.update(
+                    {"hlm_zeroecc": coprecessing_zeroecc_data_dict["hlm"]})
+                # remove amplm_zeroecc, phaselm_zeroecc because these are in
+                # the inertial frame and are given priority over hlm_zeroecc
+                # when using data for egw
+                data_dict.pop("amplm_zeroecc", None)
+                data_dict.pop("phaselm_zeroecc", None)
+        # remove inertial omegalm. The coprecessing omegalm should be computed
+        # from the coprecessing phase later.
```
vijayvarma392, line 784: Print a warning that you are doing this and why.

gw_eccentricity/eccDefinition.py
```diff
@@ -654,11 +697,97 @@ def process_data_dict(self,
...
+            amplm_dict = self.get_amplm_from_dataDict(data_dict)
+            phaselm_dict = self.get_phaselm_from_dataDict(data_dict)
+            hlm_dict = {}
+            # check if "hlm_zeroecc" is not in data_dict but amplm_zeroecc is
+            # in data_dict
+            if "amplm_zeroecc" in data_dict and "hlm_zeroecc" not in data_dict:
+                add_hlm_zeroecc = True
+                hlm_zeroecc_dict = {}
+            else:
+                add_hlm_zeroecc = False
```
vijayvarma392, line 754: Again, seems very convoluted to do handle zeroecc this way. Can it be done more simply using tags like above?

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Yes, using a tag is a much better way to handle it. Done!

gw_eccentricity/eccDefinition.py
md-arif-shaikh: Added warning.

md-arif-shaikh:
I have addressed all the comments @vijayvarma392

vijayvarma392 (changes requested):

gw_eccentricity/eccDefinition.py
```diff
@@ -579,6 +619,22 @@ def process_data_dict(self,
                 "1. 'hlm' OR \n"
                 "2. 'amplm' and 'phaselm'\n"
                 "But not both 1. and 2. at the same time."))
+        # if the system is precessing and the modes are given in the inertial
+        # frame, rotate the modes and obtain the corresponding modes in the
+        # coprecessing frame
+        if self.precessing is True and self.frame == "inertial":
+            warnings.warn(
```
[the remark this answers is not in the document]
md-arif-shaikh: yes, that came to my mind as well. However, the issue was that this was happening while processing the dataDict before setting values to debug_level. The debug_level is available via extra_kwargs which we set only after processing the dataDict.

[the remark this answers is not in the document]
vijayvarma392: I see, was there a good reason for doing it like that? Would it make sense to flip things around, or even just read or pass extra_kwargs['debug_level'] where needed?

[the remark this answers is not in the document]
md-arif-shaikh: Will the following work? Inside `process_dataDict` we check if `debug_level` is provided by the user via `extra_kwargs` and use it if its there else use the default value 0 as in the `default_extra_kwargs`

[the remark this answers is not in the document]
md-arif-shaikh: The reason was that we needed information, for example, the width for peak finding, from the `process_dataDict` and then only set the default values.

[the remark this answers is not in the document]
md-arif-shaikh: Replaced `warnings.warn` with `debug_message`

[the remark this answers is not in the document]
md-arif-shaikh: Done!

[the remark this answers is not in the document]
vijayvarma392: Hmm, don't know if that's the right approach...now there are two places where debug_level has defaults being set. Would it be better to just take it out of extra_kwargs?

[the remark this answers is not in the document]
md-arif-shaikh: But if `extra_kwargs` is not provided then it's None which will raise an error.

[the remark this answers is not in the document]
vijayvarma392: No, I mean, what is the entire debug_level option is made independent of extra_kwargs, like we did for https://github.com/vijayvarma392/gw_eccentricity/blob/main/gw_eccentricity/eccDefinition.py#L144. I don't know if this is the right approach, asking.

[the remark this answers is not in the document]
md-arif-shaikh: Yes, that will solve this issue. Will need to make changes at several places though so that the checks pass. Would it be better to make that change in a separate PR?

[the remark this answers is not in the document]
vijayvarma392: Sounds good. Revert the last commit first?

md-arif-shaikh:
Is it `git commit --amend` and then  `git push --force`?

vijayvarma392:
> Is it `git commit --amend` and then `git push --force`?

Better is `git revert <git hash>`

gw_eccentricity/eccDefinition.py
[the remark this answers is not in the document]
md-arif-shaikh: Reverted.

vijayvarma392 (approved):

Issue #201, opened 2023-07-13
how to handle very long waveforms


md-arif-shaikh:
Currently, to measure eccentricity even at a single instance of time, we construct the e(t) interpolant using the full waveform. For very long waveforms, e.g. EMRI waveforms, this might be expensive.

**Possible solution**: We consider only the part of the waveform around the given reference time. We may want to use a number of orbits on either side of the reference time. We can choose the default to be the full waveform.

PR #238, opened 2025-10-22
use segment around given reference point

md-arif-shaikh:
For long waveforms, it is not necessary to use the full waveform for measuring eccentricity. Using the full waveform will increase the cost of `egw` measurement because it will have to find all the extrema in the data. To reduce this unnecessary computational cost, we can choose a certain number of orbits long data around the reference point. This PR addresses this issue #201

### DIFF
--- a/gw_eccentricity/eccDefinition.py
+++ b/gw_eccentricity/eccDefinition.py
@@ -1947,7 +1947,7 @@ class eccDefinition:
         https://github.com/vijayvarma392/gw_eccentricity/wiki/Full-waveform-vs-short-segment#reference-frequency
         """
         # The actual tref must lie between the first and the last times where 
-        # omega_gw crosses 2 * pi * fref_in. We use only the segment between
+        # omega_gw crosses 2 * pi * fref. We use only the segment between
         # the first and the last crossings to build the secular trend of 
         # omega_gw.
         omega_ref = 2 * np.pi * fref
@@ -1996,7 +1996,7 @@ class eccDefinition:
         # get the secular trend
         secular_trend = QusicircularFitForMappingFrequencyToTime(time, omega)
         res = secular_trend.least_square_fit()
-        # get the location where the secular trend crosses fref_in
+        # get the location where the secular trend crosses omega_ref
         idx = np.argmin(np.abs(secular_trend.model(time, *res.x) - omega_ref))
         return time[idx]
         


gw_eccentricity/eccDefinition.py
```diff
@@ -451,6 +481,13 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
         # compute residual data
         if "amplm_zeroecc" in self.dataDict and "omegalm_zeroecc" in self.dataDict:
             self.compute_res_amp_gw_and_res_omega_gw()
+        # use a flag so that the function
+        # `get_segment_of_data_for_finding_extrema` is called only once to get
+        # the relevant segment of data around the reference point.
+        # The function `get_segment_of_data_for_finding_extrema` is called from
+        # within the function `find_extrema`. This flag is set to False after
+        # `find_extrema` is called for the first time.
+        self.get_segment_of_data = self.extra_kwargs["use_only_these_many_orbits"]
```
vijayvarma392: Why not call this self.use_only_these_many_orbits? They are supposed to mean the same thing.

gw_eccentricity/eccDefinition.py
```diff
@@ -1806,6 +1871,319 @@ def check_if_dropped_too_many_extrema(self, original_extrema, new_extrema,
                           f"original {extrema_type} was dropped.",
                           self.debug_level, important=False)
 
+    def get_approximate_tref_for_fref_using_secular_trend(
+            self, fref):
+        """Get approximate value of tref that corresponds to a given fref.
+
```
vijayvarma392: Say something about why we need this approximate tref

gw_eccentricity/eccDefinition.py
```diff
@@ -1480,10 +1518,29 @@ def get_rational_fit_for_extrema(self, x, y, data_name=None):
...
                     debug_message(
-                        "Both numerator and denominator degrees are equal to 1 "
-                        "and cannot be lowered further.",
-                        debug_level=self.debug_level, important=False)
+                        ("Both numerator and denominator degrees are equal to 1, "
+                        "but the fit is still nonmonotonic. Can not lower the degrees further. "
+                        "This may happen when omega at the extrema is not varying much and "
+                        "using even the lowest degrees is causing spurious divergences (poles).\n"
+                        + extra_message +
+                        "In such cases, we fallback to using spline."),
```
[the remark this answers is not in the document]

gw_eccentricity/eccDefinitionUsingAmplitude.py
```diff
@@ -53,6 +53,15 @@ def find_extrema(self, extrema_type="pericenters"):
...
                             " or 'apocenters'")
+        # We need to get the relevant segment from the full data only
+        # once, whereas `find_extrema` is called twice, once for the
+        # pericenters and then for the apocenters. We handle this
+        # by setting the flag `get_segment_of_data` to False after
+        # `self.get_segment_of_data_for_finding_extrema` is called for
+        # the first time.
+        if self.get_segment_of_data:
+            self.get_segment_of_data_for_finding_extrema()
+            self.get_segment_of_data = False
```
vijayvarma392: Why not just do this immediately after the waveform is loaded? Does it need to be done in Amplitude and FrequencyFits

md-arif-shaikh: We can not do this immediately after loading the waveform because we need to know the `tref_in` or `fref_in` for selecting segment around it. So, we can do this only inside `measure_ecc` which takes `tref_in` or `fref_in` as inputs. This also means that in the derived class when we initialize `self.data_for_finding_extrema = self.get_data_for_finding_extrema()` it uses the full waveform. Therefore, we need to handle the truncation of waveform separately before its used by `find_extrema`. One potential solution would be to move `tref_in`/`fref_in` to the arguments of the base class `eccDefiniton`. Should we do that?

gw_eccentricity/utils.py
```diff
@@ -402,3 +403,39 @@ def debug_message(message, debug_level, important=True,
     if debug_level == 2:
         # raise Exception
         raise Exception(message)
+
+class SecularTrend:
```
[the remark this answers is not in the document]
md-arif-shaikh: > The name is not descriptive enough..

How about `QuasicircularFitFromEccentricFrequency`?

gw_eccentricity/eccDefinitionUsingAmplitude.py
vijayvarma392: Moving to eccDefiniton might be ok. Is there a downside to doing it that way?

gw_eccentricity/utils.py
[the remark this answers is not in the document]
vijayvarma392: Much better, but if the name can include some hint about why you need this, that might be better.

gw_eccentricity/eccDefinitionUsingAmplitude.py
md-arif-shaikh: Moved `tref_in/fref_in` to `self.__init__`. Now we get the segments in `self.__init__` only.

gw_eccentricity/eccDefinition.py
```diff
@@ -376,6 +447,14 @@ def __init__(self, dataDict, num_orbits_to_exclude_before_merger=2,
             deepcopy(extra_kwargs), self.get_default_extra_kwargs(),
             "extra_kwargs",
             "eccDefinition.get_default_extra_kwargs()")
+        self.debug_plots = self.extra_kwargs["debug_plots"]
+        # Get amp, phase and omega segments around the reference points
+        if self.extra_kwargs["use_only_these_many_orbits"]:
+            approximate_num_orbits = ((self.phase_gw[-1] - self.phase_gw[0])
```
vijayvarma392: Call this `approx_total_num_orbits`?

gw_eccentricity/eccDefinition.py
```diff
@@ -1480,10 +1560,29 @@ def get_rational_fit_for_extrema(self, x, y, data_name=None):
                     self.special_interp_kwargs_for_omega_gw_extrema["denom_degree"] -= 1
                 if self.special_interp_kwargs_for_omega_gw_extrema["num_degree"] == 1 \
                     and self.special_interp_kwargs_for_omega_gw_extrema["denom_degree"] == 1:
+                    if (self.extra_kwargs["use_only_these_many_orbits"]
```
vijayvarma392: we may want `self.extra_kwargs["use_only_these_many_orbits"] is not None` here to avoid confusion?

gw_eccentricity/eccDefinition.py
```diff
@@ -1806,10 +1913,309 @@ def check_if_dropped_too_many_extrema(self, original_extrema, new_extrema,
...
+        # if omega_gw is > omega_ref, then right_indices will be
+        # empty. It implies that the fref, where we want to measure
+        # the eccentricity, is too low. That is, fref is smaller than
+        # the frequency where we can measure eccentricity.
+        if len(right_indices) == 0:
+            left = left_indices[0]
+            right = left
+        # Similarly, if fref is too high, that is, omega_gw <
+        # omega_ref, then left_indices will be empty.
+        elif len(left_indices) == 0:
```
vijayvarma392: I'm not sure why you need these if-else conditions, add a comment to explain?

vijayvarma392 (changes requested):

PR #246, opened 2026-03-24
add submodule for postprocessing

md-arif-shaikh:
Mains files to review:
- core.py
  - Contains core stateless functions and dataclasses to be used by both PostProcess class and MPI based postprocessing
- postprocess.py
  - Base class for postprocessing on local machine, for example from a notebook.
  - Designed in a way to make it easier for working with posterior generated by any PE pipeline and not only Bilby. Keeping in mind that we may test it also for LISA
- postprocess_for_bilby.py
  - Derived class PostProcessForBilby for working with Bilby posterior
- postprocess_cli.py
  - Script to perfrom postprocessing on cluster using MPI

Example dir contains notebooks and scripts to showcase postprocessing Bilby posterior.


gw_eccentricity/postprocess/examples/postprocess_for_bilby.ipynb
vijayvarma392, line 1: Should the examples live with the other examples [here](https://github.com/vijayvarma392/gw_eccentricity/tree/posterior/examples)? Otherwise, anyone that installs gwecc will get these examples in their python dirs. I'm thinking a `postprocessing` directory in examples, along with a mention of this and what it does in the main Readme.

vijayvarma392: In the example notebooks, it looks like there are huge outputs of warnings. We need to say something in the notebooks about this: what should people make of them, should they ignore them, etc.

examples/postprocessing/teobresums_dali.py
```diff
@@ -0,0 +1,109 @@
+import numpy as np
```
vijayvarma392, line 11: Explain why this file exists at the top?

gw_eccentricity/postprocess/core.py
```diff
@@ -0,0 +1,262 @@
...
+        fref: float,
+        data_dict_generator: callable,
+        data_dict_generator_extra_kwargs: dict | None = None,
+        method: str = "Amplitude",
+        gw_eccentricity_kwargs: dict | None = None) -> PostProcessResult:
+    """Measure eccentricity and mean anomaly from waveform modes for a sample.
+
+    A wrapper around ``gw_eccentricity.measure_eccentricity`` to measure
+    eccentricity from the waveform modes for a sample with given ``params``.
+
```
vijayvarma392, line 215: Is this the main function the user sees?
Whichever function that is, could you describe the entire procedure in the docs? Like how the frequency ranges are chosen, etc.

vijayvarma392 (changes requested):

gw_eccentricity/postprocess/examples/postprocess_for_bilby.ipynb
md-arif-shaikh: Yes, agree. Moved the examples to top level examples/postprocessing. Added a subsection in the Usage section in the main Readme.

Yes, I have cleaned up the notebook. Have another look, it looks much better. Earlier there were two sources for long outputs -- one was because I was using posterior for lower eccentricity and some of the samples were failing (because I was using Amplitude). Others were coming from rational_fit. Changed both and the notebook looks much more cleaner now.

examples/postprocessing/teobresums_dali.py
md-arif-shaikh: Done. Also renamed the file.

gw_eccentricity/postprocess/core.py
md-arif-shaikh: Okay. Shall I expand the docs for the params? or are you suggesting for a detailed description at the top of the function docstring?

md-arif-shaikh: Added some details on the range of allowed tref/fref at the top.

md-arif-shaikh:
@vijayvarma392 could you have another look?

vijayvarma392 (approved):

